PR #12, opened 2019-03-19
Set the user-agent string for BQ storage.

kmjung:
This change modifies the build process to automatically generate build-time configuration
constants such as project name and version, and uses them to generate a user-agent string
when constructing a new BigQuery storage client object.

### DIFF
--- a/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
+++ b/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
@@ -143,9 +143,6 @@ private[bigquery] class DirectBigQueryRelation(
 }
 
 object DirectBigQueryRelation {
-  def createHeaderProvider(): HeaderProvider = {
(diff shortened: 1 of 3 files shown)


kmjung:
R: @pmkc

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -142,11 +143,20 @@ private[bigquery] class DirectBigQueryRelation(
 }
 
 object DirectBigQueryRelation {
+  def createHeaderProvider(): HeaderProvider = {
+    FixedHeaderProvider.create("user-agent", BuildInfo.name + "/" + BuildInfo.version)
+  }
```
pmkc: I would inline a one liner only used once in Scala

kmjung: Done.

PR #24, opened 2019-06-25
Add example querying from BigQuery and loading the results into Spark

pmkc:
Caveat this approach currently only supports results of 10MB or more. Users would have to manually save smaller queries to a temp Dataset possibly with a TTL.

I think we could add this functionality to the connector without too much work. Maybe we should file a PR, but either way the documentation works.

### DIFF
--- a/examples/python/query_results.py
+++ b/examples/python/query_results.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


examples/python/query_results.py
```diff
@@ -0,0 +1,52 @@
...
+
+spark = SparkSession.builder.appName('Query Results').getOrCreate()
+bq = bigquery.Client()
+
+print('Querying BigQuery')
+query_job = bq.query(QUERY)
+
+# There is a small race condition here where the Storage API might not be able to
+# find the table right away
+time.sleep(1)
```
tswast: You might want to wait for the job to complete with `query_job.result()`. You can specify a timeout of 1 seconds if you do want to attempt reading after waiting "long enough".

pmkc: Thanks fixed!

Issue #6, opened 2019-03-04
Support specifying auth through Spark Configuration


pmkc:

FurcyPin:
@davidrabinowitz 
Could you please make a release that include this PR?

PR #27, opened 2019-06-26
BigQueryOptions default instance should not be required #6

xjrk58:
BigQueryOptions default instance should not be required outside GCE 

Remove Scalalogging that causes significant dependency issues on Databricks runtime.


davidrabinowitz:
Fixed by PR #33

PR #28, opened 2019-06-26
Replace the manual skewing code with the BALANCED sharding strategy

aryann:
The BALANCED sharding strategy ensures balance among many streams on the
server-side, obviating the need for the client to mimic such behavior. Note that
I'm using an unknown field to force the BALANCED sharding strategy. We are in
the process of releasing new Java libraries. Once the new libraries are out, I
will send another pull request to use the generated code for setting the
sharding strategy.

### DIFF
--- a/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
+++ b/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
@@ -32,7 +32,6 @@ case class SparkBigQueryOptions(
     credentialsFile: Option[String] = None,
     filter: Option[String] = None,
     schema: Option[StructType] = None,
-    skewLimit: Double = SparkBigQueryOptions.SKEW_LIMIT_DEFAULT,
     parallelism: Option[Int] = None) {
 
   def createCredentials: Option[Credentials] =
@@ -51,8 +50,6 @@ case class SparkBigQueryOptions(
 
 /** Resolvers for {@link SparkBigQueryOptions} */
 object SparkBigQueryOptions {
-  private val SKEW_LIMIT_KEY = "skewLimit"
-  private val SKEW_LIMIT_DEFAULT = 1.5
 
   def apply(
       parameters: Map[String, String],
@@ -71,13 +68,9 @@ object SparkBigQueryOptions {
       defaultBilledProject)
     val filter = getOption(parameters, "filter")
     val parallelism = getOption(parameters, "parallelism").map(_.toInt)
-    val skewLimit = getOption(parameters, SKEW_LIMIT_KEY).map(_.toDouble)
-        .getOrElse(SKEW_LIMIT_DEFAULT)
-    // BigQuery will actually error if we close the last stream.
-    assert(skewLimit >= 1.0,
-      s"Paramater '$SKEW_LIMIT_KEY' must be at least 1.0 to read all data '$skewLimit'")
+
     SparkBigQueryOptions(tableId, parentProject, credsParam, credsFileParam,
-      filter, schema, skewLimit, parallelism)
+      filter, schema, parallelism)
   }
 
 
--- a/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
+++ b/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
@@ -26,6 +26,7 @@ import com.google.cloud.bigquery.storage.v1beta1.Storage.{CreateReadSessionReque
 import com.google.cloud.bigquery.storage.v1beta1.TableReferenceProto.TableReference
 import com.google.cloud.bigquery.{Schema, StandardTableDefinition, TableDefinition, TableInfo}
 import com.google.cloud.spark.bigquery.{BigQueryRelation, BuildInfo, SparkBigQueryOptions}
+import com.google.protobuf.UnknownFieldSet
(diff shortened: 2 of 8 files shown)


aryann:
@pmkc please take a look when you get a chance!

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -112,12 +115,12 @@ private[bigquery] class DirectBigQueryRelation(
   }
 
   /**
-   * The theoretical number of Partitions of the returned DataFrame.
-   * If the table is small the server will provide fewer readers and there will be fewer
-   * partitions. If all partitions are not read concurrently, some Partitions will be empty.
-   */
+    * The theoretical number of Partitions of the returned DataFrame.
```
pmkc: nit: on formatting. I don't really care about 2 vs 4 spaces, but I try to follow Databricks style convention of javadoc style instead of scaladoc style (it's a check box in intellij scala style).

You can run sbt scalastyle and sbt test:scalastyle to check. I used to have them block the build, but they blocked compiler messages.

src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndITSuite.scala
```diff
@@ -159,25 +159,25 @@ class SparkBigQueryEndToEndITSuite extends FunSuite
...
-      // In this case it is usually 2X (booleans can be read very fast). In some rare cases
-      // this can grow by a lot thus we assert that we see less than 10X desired.
-      assert(sizeOfFirstPartition < (LARGE_TABLE_NUM_ROWS * 10.0 / df.rdd.getNumPartitions))
+
+      // Since we are only reading from a single stream, we can expect to get at least as many rows
+      // in that stream as a perfectly uniform distribution would command. Note that the assertion
+      // is on a range of rows because rows are assigned to streams on the server-side in
+      // indivisible units of many rows.
+      val numRowsLowerBound = LARGE_TABLE_NUM_ROWS / df.rdd.getNumPartitions
+      assert(numRowsLowerBound to (numRowsLowerBound * 1.1).toInt contains sizeOfFirstPartition)
```
pmkc: enumerating the values seems sub-optimal from a algorithmic perspective although I realize it's not that big. Could you use comparators?

src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -32,7 +32,6 @@ case class SparkBigQueryOptions(
     credentialsFile: Option[String] = None,
     filter: Option[String] = None,
     schema: Option[StructType] = None,
-    skewLimit: Double = SparkBigQueryOptions.SKEW_LIMIT_DEFAULT,
```
pmkc: Update README.md to document this and add a line to CHANGES.md.

Since we are removing a property this could be viewed as a breaking change, but it shouldn't break users if they continue to specify it so I don't really care about changing the version.

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -87,14 +88,16 @@ private[bigquery] class DirectBigQueryRelation(
           .setRequestedStreams(getNumPartitions)
           .setReadOptions(readOptions)
           .setTableReference(tableReference)
+          // TODO(aryann): Once we rebuild the generated client code, we should change this to
+          // use setShardingStrategy().
```
pmkc: Are you sure this won't get stripped out for less authorized users?

Are you sure this isn't released yet? We could try updating the dependencies. 

Please document a little clearer it's more than a little scary as is.

Are you planning to support not balanced going forward? It might be nice to give users the choice (and still turn it on by default)

aryann: Summarizing our earlier conversation for anyone following along: this is just a client-side issues because the generated clients haven't been updated. The BALANCED strategy is public and anyone who knows how to specify it on the wire can use it. I think in most cases, you want BALANCED. LIQUID is not particularly useful as most data processing systems want bounded sources.

src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndITSuite.scala
aryann: Good point! I thought there was some magic involved here where Scala would do something smart, but I believe that's nonsense since the param gets evaluated first.

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
aryann: I reconfigured my IntelliJ to use JavaDoc style and changed continuations to 4.

sbt scalastyle and sbt test:scalastyle don't report any problems. In some places, I also fixed continuations.

Maybe we can figure out a way to share an IntelliJ style config or have a binary style things automatically. Apache Beam uses spotless, which works pretty well.

src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
aryann: Done.

PR #29, opened 2019-06-28
Update the BigQuery Storage API client

aryann:
As part of this change, we can now call setShardingStrategy() instead of relying on the unknown fields proto API for setting the sharding strategy.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,11 +54,10 @@ val relocationPrefix = s"$myPackage.repackaged"
 
 libraryDependencies ++= Seq(
   // Keep com.google.cloud dependencies in sync
-  "com.google.cloud" % "google-cloud-bigquery" % "1.61.0",
-  // Note BigQuery Storage Generated Library is Alpha but the API is Beta
-  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.88.0-alpha",
+  "com.google.cloud" % "google-cloud-bigquery" % "1.82.0",
+  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.98.0-beta",
   // Keep in sync with com.google.cloud
-  "io.grpc" % "grpc-netty-shaded" % "1.17.1",
+  "io.grpc" % "grpc-netty-shaded" % "1.22.1",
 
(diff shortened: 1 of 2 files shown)


aryann:
@pmkc

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -88,12 +87,7 @@ private[bigquery] class DirectBigQueryRelation(
             .setRequestedStreams(getNumPartitions)
             .setReadOptions(readOptions)
             .setTableReference(tableReference)
-            // TODO(aryann): Once we rebuild the generated client code, we should change
-            // setUnknownFields() to use setShardingStrategy(ShardingStrategy.BALANCED). The
-            // BALANCED strategy is public at the moment, so setting it using the unknown fields
-            // API is safe.
```
pmkc: Some comment about what balanced means (link to any public docs?) would still be good

build.sbt
```diff
@@ -55,8 +55,8 @@ val relocationPrefix = s"$myPackage.repackaged"
 libraryDependencies ++= Seq(
   // Keep com.google.cloud dependencies in sync
   "com.google.cloud" % "google-cloud-bigquery" % "1.61.0",
```
pmkc: Please update this to keep in in sync

build.sbt
```diff
@@ -55,8 +55,8 @@ val relocationPrefix = s"$myPackage.repackaged"
...
-  // Note BigQuery Storage Generated Library is Alpha but the API is Beta
-  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.88.0-alpha",
+
+  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.98.0-beta",
   // Keep in sync with com.google.cloud
   "io.grpc" % "grpc-netty-shaded" % "1.17.1",
```
pmkc: This too

aryann:
@pmkc Please take another look. I will send out a new PR for the -SNAPSHOT change shortly.

PR #38, opened 2019-07-15
Bump the version

aryann:
This is in preparation of changed coming in from https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/29.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 lazy val commonSettings = Seq(
   organization := "com.google.cloud.spark",
   name := "spark-bigquery",
-  version := "0.7.0-beta",
+  version := "0.7.1-beta-SNAPSHOT",
   scalaVersion := "2.11.8",
   crossScalaVersions := Seq("2.10.5"),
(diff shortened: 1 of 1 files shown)


aryann:
@pmkc

build.sbt
```diff
@@ -1,7 +1,7 @@
 lazy val commonSettings = Seq(
   organization := "com.google.cloud.spark",
   name := "spark-bigquery",
-  version := "0.7.0-beta",
+  version := "0.8.0-beta-SNAPSHOT",
```
pmkc: Usually you mark the next change 0.7.1-SNAPSHOT.

If you introduce a change not worthy of a subminor I would change from 0.7.1-SNAPSHOT to 0.8.0-SHAPSHOT in that PR.

I think #29 is not big enough to merit a minor change (especially before 1.0 where minor is the new major if that makes sense).

aryann: Done. Changed to 0.7.1-SNAPSHOT.

PR #39, opened 2019-07-16
Update filtering support.

kmjung:
This change modifies the filtering support in the direct BigQuery
relation to reflect the full set of filters which are supported by
the storage API. Notably, this includes:
- Support for multi-clause predicates (e.g. AND, OR, and NOT).
- Support for pseudo-column variables (_PARTITIONDATE) for ingestion
  time partitioned tables.
- Support for additional filter operations (IsNull, IsNotNull, In,
  StringStartsWith, StringEndsWith, StringContains).

### DIFF
--- a/src/test/scala/com/google/cloud/spark/bigquery/DirectBigQueryRelationSuite.scala
+++ b/src/test/scala/com/google/cloud/spark/bigquery/DirectBigQueryRelationSuite.scala
@@ -79,35 +79,39 @@ class DirectBigQueryRelationSuite
 
   test("valid filters") {
     val validFilters = Seq(
-      GreaterThan("foo", "aardvark"),
       EqualTo("foo", "manatee"),
+      GreaterThan("foo", "aardvark"),
+      GreaterThanOrEqual("bar", 2),
       LessThan("foo", "zebra"),
       LessThanOrEqual("bar", 1),
-      GreaterThanOrEqual("bar", 2)
+      In("foo", Array(1, 2, 3)),
+      IsNull("foo"),
+      IsNotNull("foo"),
+      And(IsNull("foo"), IsNotNull("bar")),
+      Or(IsNull("foo"), IsNotNull("foo")),
+      Not(IsNull("foo")),
+      StringStartsWith("foo", "abc"),
+      StringEndsWith("foo", "def"),
+      StringContains("foo", "abcdef")
     )
     validFilters.foreach { f =>
       assert(bigQueryRelation.unhandledFilters(Array(f)).isEmpty)
     }
   }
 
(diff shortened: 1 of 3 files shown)


kmjung:
R: @cyxxy @pmkc @aniket486

src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndITSuite.scala
```diff
@@ -96,14 +95,49 @@ class SparkBigQueryEndToEndITSuite extends FunSuite
     spark.read.format("bigquery").option("table", SHAKESPEARE_TABLE).load()
   }
 
+  val filterData = Table(
+    ("condition", "elements"),
```
pmkc: nit: I think
Table(
    "condition" -> "elements",
  ...
is pretty syntactic sugar, but the official docs don't use it so either works for me.
http://www.scalatest.org/user_guide/table_driven_property_checks

src/test/scala/com/google/cloud/spark/bigquery/DirectBigQueryRelationSuite.scala
```diff
@@ -79,35 +79,39 @@ class DirectBigQueryRelationSuite
 
   test("valid filters") {
     val validFilters = Seq(
-      GreaterThan("foo", "aardvark"),
       EqualTo("foo", "manatee"),
+      GreaterThan("foo", "aardvark"),
+      GreaterThanOrEqual("bar", 2),
       LessThan("foo", "zebra"),
       LessThanOrEqual("bar", 1),
```
pmkc: I know I wrote this, but it seems weird to test type mismatches like this are valid even if they are

src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndITSuite.scala
kmjung: I'd like to keep this in sync with the official docs -- makes it easier for people to cross-reference the two.

src/test/scala/com/google/cloud/spark/bigquery/DirectBigQueryRelationSuite.scala
kmjung: Does Spark know at compile time what the types of the various columns are? Would such a filter ever be provided to the DirectBigQueryRelation, in other words?

digger:
Hi @kmjung,
Shouldn't the readme be updated accordingly? Currently it still says the connector pushes done 1 filter only.

PR #43, opened 2019-07-31
Setup presubmit and post merge tests using cloud build

aniket486:
Upgrade sbt version to workaround problem described in: https://github.com/sbt/sbt-launcher-package/pull/158/files

### DIFF
--- a/cloudbuild/Dockerfile
+++ b/cloudbuild/Dockerfile
@@ -1,0 +1,14 @@
+# This Dockerfile creates an image for running presubmit tests.
+FROM openjdk:8
+
+# This should match with sbt version under project/build.properties
+ARG SBT_VERSION=0.13.17
+
+# Install sbt
+RUN \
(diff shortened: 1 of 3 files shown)


cloudbuild/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+# This should match with sbt version under project/build.properties
+ARG SBT_VERSION=0.13.17
+
+# Install sbt
+RUN \
+  curl -L -o sbt-$SBT_VERSION.deb http://dl.bintray.com/sbt/debian/sbt-$SBT_VERSION.deb && \
+  dpkg -i sbt-$SBT_VERSION.deb && \
+  rm sbt-$SBT_VERSION.deb && \
+  apt-get update && \
+  apt-get install sbt && \
```
medb: Why do you need to install sbt with apt-get, if it already was installed by dpkg command?

aniket486: IIUC, this will ensure that apt will install any missing dependencies.

PR #45, opened 2019-08-07
Add example of configuring Jupyter notebook to use BigQuery connector

pmkc:
For some reason github isn't rendering this in my repo...


pmkc:
ACK we should do that too.

PR #49, opened 2019-08-12
Switch requested partitions from SparkContext.defaultParallelism to o…

### LINKED ISSUE #48: SparkContext.defaultParallelism is not great with dynamic allocation
Google Cloud Dataproc uses Spark's dynamic allocation by default which causes very few partitions to be requested relative to the size of the cluster.

In contrast the MapReduce BigQuery connector and Reading from GCS create one partition per ~128MB.

I did not write the Spark connector to do this, because it did not behave well with the initial sharding strategy. With the balanced Sharding strategy we should switch to that model

pmkc:
…ne partition per 400MB.

This should work better with balanced sharding and dynamic allocation.

I believe 400MB is the correct number based on the API internals, but we could lower it to 128MB to be consistent with GCS and just let the API correct it.

I also add pretty scary log spam about the API decreasing the number of partitions, but I think this will give users clarity.

Fixes #48.

### DIFF
--- a/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
+++ b/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
@@ -20,10 +20,10 @@ import java.sql.{Date, Timestamp}
 import com.google.api.gax.core.CredentialsProvider
 import com.google.api.gax.rpc.FixedHeaderProvider
 import com.google.auth.Credentials
-import com.google.cloud.bigquery.storage.v1beta1.{BigQueryStorageClient, BigQueryStorageSettings}
 import com.google.cloud.bigquery.storage.v1beta1.ReadOptions.TableReadOptions
 import com.google.cloud.bigquery.storage.v1beta1.Storage.{CreateReadSessionRequest, DataFormat, ShardingStrategy}
 import com.google.cloud.bigquery.storage.v1beta1.TableReferenceProto.TableReference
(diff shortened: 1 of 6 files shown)


kmjung:
LGTM. This approach is probably better than the current approach, but it will mean that your estimation may be (significantly) inaccurate in the presence of column selection. Unfortunately, there's no good workaround for this at the moment.

pmkc:
@aniket486 Does this look good to you?

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -53,6 +53,13 @@ private[bigquery] class DirectBigQueryRelation(
   }
   private val log = Logger(getClass)
 
+  /**
+   * Default parallelism to 1 reader per 400MB, which should be about the maximum allowed by the
+   * BigQuery Storage API. The number of partitions returned may be significantly less depending
+   * on a number of factors.
+   */
+  val DEFAULT_BYTES_PER_PARTITION = 400L * 1000 * 1000
```
aniket486: How about making this user configurable? That way, you can provide users a better way to control parallelism for the case when they are selecting only a few columns.

aniket486:
LGTM overall, except one thing - Let's make 400MB configurable in case it makes significantly inaccurate estimation in presence of column selection.

pmkc:
It already is configurable via the parallelism flag.

I can add a new parameter for the target "partitionSize" that both control the same thing, but that flag is going to be wildly off in the normal case of column or predicate filtering.

Issue #50, opened 2019-08-16
Support partitioning pseudo columns


pmkc:
This is tricky, because AFAIK Spark has no support for pseudo columns

```
val df = spark.read.format("bigquery").option("table", TABLE) \
  .where($"_PARTITIONDATE" > "2019-01-01")
```
would require materializing _PARTITIONDATE (at least in the schema), which the BigQuery Storage API does not support.

A work around today is to use the filter option

```
val df = spark.read.format("bigquery").option("table", TABLE) \
  .option("filter", "_PARTITIONDATE > '2019-01-01'")
```

should work today

aniket486:
@davidrabinowitz we should document this in the readme.

PR #58, opened 2019-09-13
#50 Documenting querying by pseudo columns of partitioned tables

davidrabinowitz:


README.md
```diff
@@ -325,6 +325,18 @@ val otherRows = cachedDF.select("word_count")
 
 You can also manually specify the `filter` option, which will override automatic pushdown and Spark will do the rest of the filtering in the client.
 
+### Partitioned Tables
+
+The pseudo columns \_PARTITIONDATE and \_PARTITIONTIME are not part of the table schema. Therefore in order to query by the partitions of [partitioned tables](https://cloud.google.com/bigquery/docs/partitioned-tables) do not use the where() method shown above. Instead, add a filter option in the following manner:
```
aniket486: We should probably identify partitioned tables and fail the query instead of doing a full table scan (and add a opt-in option to full table scan if users must do it).

PR #66, opened 2019-10-07
Adding write support

davidrabinowitz:
This PR adds DataFrame-to-BigQuery write support. In addition, it adds a Scala 2.11 and 2.12 cross build artifacts.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,10 +1,14 @@
+lazy val scala211Version = "2.11.8"
+lazy val scala212Version = "2.12.8"
+
 lazy val commonSettings = Seq(
   organization := "com.google.cloud.spark",
   name := "spark-bigquery",
-  version := "0.8.2-beta-SNAPSHOT",
-  scalaVersion := "2.11.8",
-  crossScalaVersions := Seq("2.10.5"),
-  sparkVersion := "2.3.2",
+  version := "0.9.0-beta-SNAPSHOT",
+  
+  scalaVersion := scala211Version,
+  crossScalaVersions := Seq(scala211Version, scala212Version),
+  sparkVersion := "2.4.4",
   spName := "google/spark-bigquery",
   sparkComponents ++= Seq("core", "sql")
 )
@@ -44,7 +48,10 @@ val excludedOrgs = Seq(
   // All use jackson-core-asl:1.9.13
   "org.codehaus.jackson",
   // All use SLF4J 
-  "org.slf4j"
+  "org.slf4j",
+  "com.sun.jdmk",
+  "com.sun.jmx",
+  "javax.jms"
 )
 val myPackage = "com.google.cloud.spark.bigquery"
 
@@ -56,18 +63,18 @@ val relocationPrefix = s"$myPackage.repackaged"
 
 libraryDependencies ++= Seq(
   // Keep com.google.cloud dependencies in sync
-  "com.google.cloud" % "google-cloud-bigquery" % "1.82.0",
-  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.98.0-beta",
+  "com.google.cloud" % "google-cloud-bigquery" % "1.96.0",
+  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.114.0-beta",
   // Keep in sync with com.google.cloud
-  "io.grpc" % "grpc-netty-shaded" % "1.22.1",
+  "io.grpc" % "grpc-netty-shaded" % "1.23.0",
+  "com.google.guava" % "guava" % "28.1-jre",
 
   "org.apache.avro" % "avro" % "1.8.2",
 
   // Brings SLF4J 1.7.25. Spark 2.2 has 1.7.16
-  // TODO(pmkc): Shade
   "com.typesafe.scala-logging" %% "scala-logging" % "3.7.2",
 
-
+  "com.google.cloud.bigdataoss" % "gcs-connector" % "hadoop2-2.0.0" % "test",
   "org.scalatest" %% "scalatest" % "3.0.4" % "test",
   "org.mockito" % "mockito-all" % "1.10.19" % "test")
     .map(_.excludeAll(excludedOrgs.map(ExclusionRule(_)): _*))
--- a/build.sbt
+++ b/build.sbt
@@ -74,7 +74,10 @@ libraryDependencies ++= Seq(
   // Brings SLF4J 1.7.25. Spark 2.2 has 1.7.16
   "com.typesafe.scala-logging" %% "scala-logging" % "3.7.2",
 
-  "com.google.cloud.bigdataoss" % "gcs-connector" % "hadoop2-2.0.0" % "test",
+  // runtime
+  "com.google.cloud.bigdataoss" % "gcs-connector" % "hadoop2-2.0.0" % "runtime",
+
+  // test
   "org.scalatest" %% "scalatest" % "3.0.4" % "test",
   "org.mockito" % "mockito-all" % "1.10.19" % "test")
     .map(_.excludeAll(excludedOrgs.map(ExclusionRule(_)): _*))
--- a/src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
+++ b/src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
@@ -1,0 +1,61 @@
+package com.google.cloud.spark.bigquery
+
+import java.math.BigInteger
+
+import com.google.cloud.bigquery.{BigQuery, TableDefinition}
+import com.typesafe.scalalogging.StrictLogging
+import org.apache.spark.sql.sources.{BaseRelation, InsertableRelation}
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.{DataFrame, SQLContext, SaveMode}
+
+// Used only to insert data into BigQuery. Getting the table metadata (existence,
+// number of records, size, etc.) is done by API calls without reading existing
+// data
+case class BigQueryInsertableRelation(val bigQuery: BigQuery,
+                                      val sqlContext: SQLContext,
+                                      val options: SparkBigQueryOptions)
+  extends BaseRelation
+    with InsertableRelation
+    with StrictLogging {
+
+
+  override def insert(data: DataFrame, overwrite: Boolean): Unit = {
+    logger.info(s"insert data=${data}, overwrite=$overwrite")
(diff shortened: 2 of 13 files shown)


build.sbt
```diff
@@ -56,18 +63,18 @@ val relocationPrefix = s"$myPackage.repackaged"
...
+  "com.google.guava" % "guava" % "28.1-jre",
 
   "org.apache.avro" % "avro" % "1.8.2",
 
   // Brings SLF4J 1.7.25. Spark 2.2 has 1.7.16
-  // TODO(pmkc): Shade
   "com.typesafe.scala-logging" %% "scala-logging" % "3.7.2",
 
-
+  "com.google.cloud.bigdataoss" % "gcs-connector" % "hadoop2-2.0.0" % "test",
```
medb: Will it be enough `gcsio` or `google-cloud-storage` dependencies for tests? `gcs-connector` pretty heavy-weight because it brings Hadoop with it.

davidrabinowitz: As I'm using the HDFS API to access GCS (needed for the Spark parquet/orc writing), I've changed the scope to "runtime", and added a verification that the GCS connector is installed.

src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  override def insert(data: DataFrame, overwrite: Boolean): Unit = {
+    logger.info(s"insert data=${data}, overwrite=$overwrite")
+    // the helper also supports the v2 api
+    val saveMode = if (overwrite) SaveMode.Overwrite else SaveMode.Append
+    val helper = BigQueryWriteHelper(bigQuery, sqlContext, saveMode, options, data)
+    helper.writeDataFrameToBigQuery
+  }
+
+  /**
```
pmkc: run sbt scalastyle
databricks styleguide says always use javadoc comment

src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
```diff
@@ -0,0 +1,61 @@
...
+  def exists: Boolean = {
+    // See https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/bigquery/BigQuery.html#getTable-com.google.cloud.bigquery.TableId-com.google.cloud.bigquery.BigQuery.TableOption...-
+    val table = getTable
+    table.isDefined
+  }
+
+  /**
+    * Is this table empty? A none-existing table is considered to be empty
+    */
+  def isEmpty: Boolean = numberOfRows.map(n => n.longValue() == 0).getOrElse(true)
```
pmkc: nit: don't define and test public helpers that are used in only one place

src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.{DataFrame, SQLContext, SaveMode}
+
+// Used only to insert data into BigQuery. Getting the table metadata (existence,
+// number of records, size, etc.) is done by API calls without reading existing
+// data
+case class BigQueryInsertableRelation(val bigQuery: BigQuery,
+                                      val sqlContext: SQLContext,
+                                      val options: SparkBigQueryOptions)
+  extends BaseRelation
```
pmkc: Why not check if the table exists in relationprovider and inherit from bigqueryrelation?

It seems like you are unnecessarily duplicating code

src/main/scala/com/google/cloud/spark/bigquery/BigQueryWriteHelper.scala
```diff
@@ -0,0 +1,101 @@
...
+    val fs = gcsPath.getFileSystem(conf)
+    if (fs.exists(gcsPath)) {
+      throw new IllegalStateException(s"Conflict occurred creating export directory. Path %gcsPath already exists")
+    }
+    gcsPath
+  }
+
+  def writeDataFrameToBigQuery : Unit = {
+    try {
+      val format = options.intermediateFormat.getType.toLowerCase
```
pmkc: it's hacky but I would add cleanTemporaryGcsPath as a shutdownhook in the JVM (SparkContext has a method for it) so that if the user kills their application the data is in theory cleaned up.

src/main/scala/com/google/cloud/spark/bigquery/BigQueryWriteHelper.scala
```diff
@@ -0,0 +1,101 @@
...
+  def cleanTempraryGcsPath: Unit = {
+    //TODO(davidrab): add flag to disable the deletion?
+    val fs = temporaryGcsPath.getFileSystem(conf)
+    fs.delete(temporaryGcsPath, true)
+  }
+
+
+}
+
+case class ToIterator[E](remote:RemoteIterator[E]) extends Iterator[E] {
```
pmkc: nit: I would nest this

src/main/scala/com/google/cloud/spark/bigquery/BigQueryWriteHelper.scala
```diff
@@ -0,0 +1,101 @@
...
+                               data: DataFrame) extends StrictLogging {
+
+  import org.apache.spark.sql.SaveMode
+
+  val conf = sqlContext.sparkSession.sparkContext.hadoopConfiguration
+
+  val temporaryGcsPath = {
+
+    val gcsPathOption = options.temporaryGcsBucket.map(bucket =>
+      s"gs://$bucket/.spark-bigquery-${System.currentTimeMillis()}-${UUID.randomUUID()}")
```
pmkc: nit: It's usually friendlier to use the spark appid or some such instead of currenttimemillis
You can find an id in SparkEnv.get

build.sbt
pmkc: I would not list it as runtime OR validate it. Just leave it test and let it fail.

"no filesystem for scheme gs" is a clean failure and it should fail early enough that you don't mutate any state.

src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
davidrabinowitz: BigQueryRelation holds a reference to a TableInfo member. When the table is inserted in many cases the table does not exist, so we cannot have an instance of this member (generated by the BQ client). Notice the different implementation and return value of getTable(), and the different implementation of schema()

src/main/scala/com/google/cloud/spark/bigquery/BigQueryWriteHelper.scala
davidrabinowitz: Added. Notice that the ShutdownHookManager is a Spark private API (See https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/ShutdownHookManager.scala), so I've removed this condition from the scalastyle-config file. Changed in https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/66/commits/92d7648d9edfb859a70dc4048453fd3fdae73160

src/main/scala/com/google/cloud/spark/bigquery/BigQueryWriteHelper.scala
davidrabinowitz: Fixed in https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/66/commits/ff58be6fa9b07dba8e024e7d50c540813a27c8ee

src/main/scala/com/google/cloud/spark/bigquery/BigQueryWriteHelper.scala
davidrabinowitz: Fixed in https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/66/commits/9444852da1aa2884bbf922b04e10aa93a2008d4e

src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
davidrabinowitz: Fixed

src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
davidrabinowitz: Fixed in https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/66/commits/5b5d2c14c49e76e857bbe1019ae06d51cd811720

build.sbt
davidrabinowitz: Makes sense. Fixed in https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/66/commits/5b5d2c14c49e76e857bbe1019ae06d51cd811720

src/main/scala/com/google/cloud/spark/bigquery/BigQueryInsertableRelation.scala
pmkc: You could refactor that. Or have your own TableDefinition that is just a wrapper of TableId (and maybe schema):
TableInfo.newBuilder(TableId.of("foo","bar"), StandardTableDefinition.newBuilder().build).build

but I suppose the duplication is ok

src/main/scala/com/google/cloud/spark/bigquery/BigQueryWriteHelper.scala
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+/**
+ * Resposible for recursively deleting an HDFS path. Implementing Thread in order
+ * to act as shutdown hook
+ * @param path the path to delete
+ * @param conf the hadoop configuration
+ */
+case class HdfsPathDeleter(path: Path, conf: Configuration) extends Thread with StrictLogging {
```
pmkc: nit: never call GCS HDFS.
It is either HCFS (compatible) or just filesystem

In this case I might say IntermediateDataCleaner or something

davidrabinowitz: Fixed in https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/66/commits/d34dc2094abdee77ad48558ebe9490e0ff901896

PR #68, opened 2019-10-11
Minor fixes

davidrabinowitz:
- A NPE in the shutdown hook has occurred in case the delete had succeeded in the first time. The method now verifies the path exists before trying to delete it.
- Added support for data.write.bigquery("table") implicit import, fixed regression caused by relying of shaded scalalogging

### DIFF
--- a/src/main/scala/com/google/cloud/spark/bigquery/examples/Shakespeare.scala
+++ b/src/main/scala/com/google/cloud/spark/bigquery/examples/Shakespeare.scala
@@ -26,7 +26,7 @@ object Shakespeare {
 
     // Use the Cloud Storage bucket for temporary BigQuery export data used
     // by the connector. This assumes the Cloud Storage connector for
-    // Hadoop is configured.r
+    // Hadoop is configured.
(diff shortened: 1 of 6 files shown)


src/main/scala/com/google/cloud/spark/bigquery/examples/Shakespeare.scala
```diff
@@ -15,23 +15,32 @@
...
   def main(args: Array[String]) {
-    val spark = SparkSession.builder().appName("test").getOrCreate()
-    val sc = spark.sparkContext
+    val spark = SparkSession.builder()
+      .appName("spark-bigquery-demo")
+      .getOrCreate()
+
+    // Use the Cloud Storage bucket for temporary BigQuery export data used
+    // by the connector. This assumes the Cloud Storage connector for
+    // Hadoop is configured.r
```
cyxxy: delete the trailing "r"?

davidrabinowitz: oops - typo

PR #75, opened 2019-10-29
Fix filter when sending load job to Big Query

TheForgotten69:
Currently, when writing delta tables we have _committed _started _SUCCESS file. So we need to filter only using the temporary format


TheForgotten69:
@googlebot I signed it!

davidrabinowitz:
Thanks for the patch! One comment though - can you please convert the format variable to formatSuffix, by moving the `s".$format"` outside the lambda?

Issue #22, opened 2019-06-13
Possibility to change grpc message size for a client


aborgatin:
Sometimes we got following error in this connector:

> io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Sent message larger than max (10485777 vs. 10485760)

It would be good if library client will able to change grpc message size. 
For now message size is standard - 10 * 1024 * 1024

achelimed:
Hi,
My spark job worked fine until oct, 24 when hitting the following error:
`com.google.cloud.spark.bigquery.io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: there was an error operating on 'projects/**********/locations/us/streams/EghHYzKJHGSKHYyejoCaXIoAQ': a single table row is larger than the maximum message size`

However, there was no change neither from BQ side nor the spark job code.
Any help is appreciated.

Thanks

coryfoo:
@kmjung I have filed an [issue](https://issuetracker.google.com/issues/143730055) in the tracker just now that has our unredacted project stream name.  Hopefully that helps you all get to the bottom of it.

davidrabinowitz:
Fixed in 0.9.2-beta

Issue #74, opened 2019-10-25
.option("filter", filter) will overwrite the pushdown filters


gengliangwang:
Currently, if we run
```
spark.read.format("bigquery").option("fitler", "_PARTITIONDATE > '2019-01-01'")....createOrReplaceTempView("t")
spark.sql("select * from t where a =1").show()
```

Then only the partition column filter is pushed down.
Have you considered combining both predicates with "AND"?

davidrabinowitz:
Fixed in 0.9.2-beta

PR #79, opened 2019-11-07
Bug fixes: Issue #22 and issue #74

davidrabinowitz:
Issue #22 - update to the latest BigQuery Storage client in order to support large rows
Issue #74 - pushing down all filters

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -62,10 +62,10 @@ val relocationPrefix = s"$myPackage.repackaged"
 
 libraryDependencies ++= Seq(
   // Keep com.google.cloud dependencies in sync
-  "com.google.cloud" % "google-cloud-bigquery" % "1.96.0",
-  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.114.0-beta",
+  "com.google.cloud" % "google-cloud-bigquery" % "1.99.0",
+  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.117.0-beta",
   // Keep in sync with com.google.cloud
-  "io.grpc" % "grpc-netty-shaded" % "1.23.0",
+  "io.grpc" % "grpc-netty-shaded" % "1.24.1",
   "com.google.guava" % "guava" % "28.1-jre",
 
   "org.apache.avro" % "avro" % "1.8.2",
--- a/src/main/java/com/google/cloud/spark/bigquery/examples/JavaShakespeare.java
+++ b/src/main/java/com/google/cloud/spark/bigquery/examples/JavaShakespeare.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.spark.bigquery.examples;
+
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+
+public class JavaShakespeare {
+
+  public static void main(String[] args) {
+    SparkSession spark = SparkSession.builder()
+        .appName("spark-bigquery-demo")
+        .getOrCreate();
+
+    // Use the Cloud Storage bucket for temporary BigQuery export data used
+    // by the connector. This assumes the Cloud Storage connector for
+    // Hadoop is configured.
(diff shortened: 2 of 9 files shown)


build.sbt
```diff
@@ -62,10 +62,10 @@ val relocationPrefix = s"$myPackage.repackaged"
 
 libraryDependencies ++= Seq(
   // Keep com.google.cloud dependencies in sync
-  "com.google.cloud" % "google-cloud-bigquery" % "1.96.0",
-  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.114.0-beta",
+  "com.google.cloud" % "google-cloud-bigquery" % "1.99.0",
+  "com.google.cloud" % "google-cloud-bigquerystorage" % "0.117.0-beta",
   // Keep in sync with com.google.cloud
```
kmjung: How should these be kept in sync? If there's a table somewhere, then can you add a link?

src/main/java/com/google/cloud/spark/bigquery/examples/JavaShakespeare.java
```diff
@@ -0,0 +1,50 @@
...
+        .getOrCreate();
+
+    // Use the Cloud Storage bucket for temporary BigQuery export data used
+    // by the connector. This assumes the Cloud Storage connector for
+    // Hadoop is configured.
+    String bucket = spark.sparkContext().hadoopConfiguration().get("fs.gs.system.bucket");
+    spark.conf().set("temporaryGcsBucket", bucket);
+
+    // Load data in from BigQuery.
+    Dataset<Row> wordsDF = spark.read().format("bigquery")
```
kmjung: Nit: do you want to add a filter in here for demonstration purposes?

src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndITSuite.scala
```diff
@@ -331,6 +331,32 @@ class SparkBigQueryEndToEndITSuite extends FunSuite
...
+      .load())
+
+    val oldBehaviourWords = extractWords(
+      spark.read.format("bigquery")
+      .option("table","publicdata.samples.shakespeare")
+      .option("filter","length(word) = 1")
+      .option("combinePushedDownFilters","false")
+      .load())
+
+    newBehaviourWords should equal (oldBehaviourWords)
```
kmjung: Does "should equal" imply ordering here? The read API does not guarantee that reading from the same table will return results in the same order.

build.sbt
pmkc: Not sure if there is a table

I use sbt dependencyTree personally, but you can link to github at the release commit

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -139,11 +139,24 @@ private[bigquery] class DirectBigQueryRelation(
...
+      )
+        .flatten
+        .map(f => s"($f)")
+        .mkString(" AND ")
+    } else {
+      // old behaviour, kept for backward compatibility
+      // If a manual filter has been specified do not push down anything.
+      options.filter.getOrElse {
+        // TODO(pclay): Figure out why there are unhandled filters after we already listed them
+        DirectBigQueryRelation.compileFilters(handledFilters(filters))
```
pmkc: I think boolean and filter should be independent.

Make the boolean "passThroughSparkFilters" and there are 4 valid cases:
passThrough & filter
passThrough & no filter
no passThrough & filter
no assThrough & no filter

admittedly the user can get the last behavior with noCombine and filter = '', but it makes more sense to me to make them independent

build.sbt
davidrabinowitz: The best place is the google-cloud-java BOM (cc) but unfortunately
a. com.google.cloud:google-cloud-bigquerystorage is missing (I've created https://github.com/googleapis/google-cloud-java/issues/6710)
b. SBT does not support BOM yet (See https://github.com/sbt/sbt/issues/4531)

Perhaps we can move to maven/gradle at some point, especially as the DataSources v2 API is java based rather than scala.

The BOM can be still used as a reference for the versions

src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndITSuite.scala
davidrabinowitz: No, the helper method collects the data and converts it to a set

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
davidrabinowitz: I may be missing something, but why are they dependent? The filter is set by the 'filter' option, and the flag is set by the 'combinePushedDownFilters' option. The flag controls how the filter option is handled when creating the filter to the BQ storage API - AND'ed with the where() filters if the flag is true, or potentially overrides them if the flag is false.

pmkc: My proposed boolean option is slightly different. It controls whether to pass through Spark Filters regardless of whether the user passes a filter.

So if the user sets the boolean to False and does not provide a filter, the connector does not pass any filters through at all, whereas in your code it would pass Spark's filters through.

Issue #21, opened 2019-06-11
Support reading from BQ Views


pupamanyu:
Currently reading from BigQuery Views results in the below exception. Views are used for abstracting the base tables and provide access controls on BigQuery. 

To replicate the issue:

Create a view on BigQuery
Use the view name as Table Name for spark.read operation

Exception:

Exception in thread "main" java.lang.RuntimeException: Table type VIEW is not supported.
        at scala.sys.package$.error(package.scala:27)
        at com.google.cloud.spark.bigquery.BigQueryRelationProvider.createRelationInternal(BigQueryRelationProvider.scala:55)
        at com.google.cloud.spark.bigquery.BigQueryRelationProvider.createRelation(BigQueryRelationProvider.scala:40)
        at org.apache.spark.sql.execution.datasources.DataSource.resolveRelation(DataSource.scala:341)
        at org.apache.spark.sql.DataFrameReader.loadV1Source(DataFrameReader.scala:239)
        at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:227)
        at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:164)
        at com.google.cloud.spark.bigquery.package$BigQueryDataFrameReader.bigquery(package.scala:24)
        at com.it.bi.etl.ATR_Repurchase$.main(ATR_Repurchase.scala:18)
        at com.it.bi.etl.ATR_Repurchase.main(ATR_Repurchase.scala)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.apache.spark.deploy.JavaMainApplication.start(SparkApplication.scala:52)
        at org.apache.spark.deploy.SparkSubmit$.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:894)
        at org.apache.spark.deploy.SparkSubmit$.doRunMain$1(SparkSubmit.scala:198)
        at org.apache.spark.deploy.SparkSubmit$.submit(SparkSubmit.scala:228)
        at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:137)
[...]

PR #83, opened 2019-11-13
Preliminary reading from views support (Issue #21)

davidrabinowitz:


src/main/scala/com/google/cloud/spark/bigquery/BigQueryUtil.scala
```diff
@@ -65,4 +66,7 @@ object BigQueryUtil {
   }
 
   def noneIfEmpty(s: String): Option[String] =  Option(s).filterNot(_.trim.isEmpty)
+
+  def convertAndThrow(error: BigQueryError) : Unit =
+    throw new BigQueryException(UNKNOWN_CODE, error.getMessage, error)
```
kmjung: Is it appropriate to always use UNKNOWN_CODE here?

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -127,23 +150,90 @@ private[bigquery] class DirectBigQueryRelation(
...
+      log.debug(s"querySql is $querySql")
+      destinationTableCache.get(querySql, DestinationTableBuilder(querySql))
+    } else {
+      // use the default one
+      table
+    }
+  }
+
+  def createTableFromQuery(querySql: String): TableInfo = {
+    val destinationTable = createDestinationTable
```
kmjung: How are these temporary tables deleted after creation?

src/main/scala/com/google/cloud/spark/bigquery/BigQueryUtil.scala
davidrabinowitz: The code is modeled after [BigQueryException.translateAndThrow()](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryException.java#L98). As [BigQueryError](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryError.java) does not contain the code, I see no other option.

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
davidrabinowitz: I thought that it was set to 24 hours automatically... The connector now updates the destination table's expiration time immediately after it is created

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -127,23 +150,97 @@ private[bigquery] class DirectBigQueryRelation(
...
+  // return empty if no filters are used
+  def createWhereClause(filters: Array[Filter]): Option[String] = {
+    val filtersString = DirectBigQueryRelation.compileFilters(filters)
+    BigQueryUtil.noneIfEmpty(filtersString)
+  }
+
+  def createDestinationTable: TableId = {
+    val uuid = UUID.randomUUID()
+    val name = s"_sbc_${uuid.getMostSignificantBits.toHexString}${uuid.getLeastSignificantBits.toHexString}"
+    TableId.of(tableId.getProject, tableId.getDataset, name)
```
kmjung: You might consider making this configurable -- in some cases, it won't be appropriate to put the temporary table in the same dataset as the source table. Beam, for example, creates not only a temporary table, but a temporary dataset as well which it deletes at the end of pipeline execution.

davidrabinowitz: Added the configuration options, see https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/83/commits/1cb67b5f60c38fd9575050b53ffd7d784dfc3b50

src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
davidrabinowitz: See https://github.com/GoogleCloudPlatform/spark-bigquery-connector/pull/83/commits/bdabb4216ea95818bf43569ad7d2e53e61d24c84

kmjung:
Thank you, David!

Issue #72, opened 2019-10-17
Change the shaded jar name from classifier to a new artifact name


davidrabinowitz:
At the moment the shaded jar is published as a classifier on the original jar (such as com.google.cloud.spark.bigquery:spark-bigquery_2.11:shaded) but unfortunately spark does not permit the usage of a classifier.

It means that the shaded version should be deployed under another artifact name.

bobingm:
I want to followup this issue. Any updates on this?

davidrabinowitz:
I'm working on it, hope to release it soon

davidrabinowitz:
`com.google.cloud.spark:spark-bigquery-with-dependencies_2.11:0.12.0-beta` and `com.google.cloud.spark:spark-bigquery-with-dependencies_2.12:0.12.0-beta` are now available

PR #104, opened 2020-01-15
Issue #72: Moved the shaded jar name from classifier to a new artifact name

davidrabinowitz:
The solution is based on the suggestion from the [sbt-assembly git repo](https://github.com/sbt/sbt-assembly/#q-despite-the-concerned-friends-i-still-want-publish-fat-jars-what-advice-do-you-have). I have now 3 sub-projects:
* connector - moved to a sub directory, contains all the sources code with almost no changes (a path in one test was changed)
* fatJar - creates the definition of the fat jar
* withDeps - the published shaded artifact


davidrabinowitz:
The solution is based on the suggestion from the [sbt-assembly git repo](https://github.com/sbt/sbt-assembly/#q-despite-the-concerned-friends-i-still-want-publish-fat-jars-what-advice-do-you-have). I have now 3 sub-projects:
* connector - moved to a sub directory, contains all the sources code with almost no changes (a path in one test was changed)
* fatJar - creates the definition of the fat jar
* withDeps - the published shaded artifact

PR #140, opened 2020-03-16
Add an iterator to support Arrow Schema for BQ -> Spark

Gaurangi94:
1. Added support for Arrow iterator
2. Refactored Avro iterator
3. Added unit test to ensure compatibility between arrow and avro


connector/src/main/scala/com/google/cloud/spark/bigquery/ArrowBinaryIterator.scala
```diff
@@ -0,0 +1,105 @@
...
+
+import collection.JavaConverters._
+
+/**
+ * An iterator for scanning over rows serialized in Arrow format
+ * @param columnsInOrder Sequence of columns in the schema
+ * @param schema Schema in avro format
+ * @param rowsInBytes Rows serialized in binary format for Arrow
+ */
+class ArrowBinaryIterator(columnsInOrder: Seq[String],
```
Gaurangi94: Make both ArrowBinaryIterator and AvroBinaryIterator inherit from a defined interface.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+
+  @Override
+  public ColumnarArray getArray(int rowId) {
+    if (isNullAt(rowId)) return null;
+    return accessor.getArray(rowId);
+  }
+
+  @Override
+  public ColumnarMap getMap(int rowId) {
+    throw new UnsupportedOperationException();
```
emkornfield: Why is this unsupported?

Gaurangi94: We do not need this function since we are only supporting types that are supported by BigQuery. 

Please note that this code has been adapted from https://github.com/apache/spark/blob/master/sql/catalyst/src/main/java/org/apache/spark/sql/vectorized/ArrowColumnVector.java#L30 and modified to make sure all data types supported by BigQuery are included.

connector/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -27,13 +27,14 @@ import com.google.cloud.bigquery.storage.v1beta1.Storage.{CreateReadSessionReque
...
 import com.google.cloud.bigquery.{BigQuery, BigQueryOptions, JobInfo, QueryJobConfiguration, Schema, StandardTableDefinition, TableDefinition, TableId, TableInfo}
-import com.google.cloud.spark.bigquery.{BigQueryRelation, BigQueryUtil, BuildInfo, SchemaConverters, SparkBigQueryOptions}
+import com.google.cloud.spark.bigquery.{BigQueryRelation, BigQueryUtil, SchemaConverters, SparkBigQueryOptions}
 import com.google.common.cache.{Cache, CacheBuilder}
 import org.apache.spark.Partition
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.sources._
+import org.json4s.BuildInfo
```
davidrabinowitz: Why do we need the json4s BuildInfo? I believe we want ours (see line 30)

Gaurangi94: fixed

connector/src/main/scala/com/google/cloud/spark/bigquery/ArrowBinaryIterator.scala
```diff
@@ -0,0 +1,105 @@
...
+    0, Long.MaxValue)
+
+  val byteStringWithSchema = schema.concat(rowsInBytes)
+
+  val arrowStreamReader = new ArrowStreamReader(
+    new ByteArrayInputStream(byteStringWithSchema.toByteArray).asInstanceOf[InputStream]
+    , allocator)
+
+  val arrowReaderIterator = new ArrowReaderIterator(arrowStreamReader)
+  val iterator = arrowReaderIterator.flatMap(root => toArrowRows(root, columnsInOrder))
```
emkornfield: I don't know if this would change your code design (you might be able to avoid an extra object allocation per row-batch which probably wouldn't change much performance wise) but root in this context will be the same object for every call (each call to loadNextBatch updates the same vectorSchemaRoot on reader).

connector/src/main/scala/com/google/cloud/spark/bigquery/AvroBinaryIterator.scala
```diff
@@ -0,0 +1,45 @@
...
+ * @param bqSchema Schema of underlying BigQuery source
+ * @param columnsInOrder Sequence of columns in the schema
+ * @param schema Schema in avro format
+ * @param rowsInBytes Rows serialized in binary format for Avro
+ */
+class AvroBinaryIterator(bqSchema: Schema,
+                         columnsInOrder: Seq[String],
+                         schema: AvroSchema,
+                         rowsInBytes: ByteString) extends Iterator[InternalRow] {
+  // TODO(pclay): replace nulls with reusable objects
```
emkornfield: is this still relevant?

connector/src/main/scala/com/google/cloud/spark/bigquery/direct/BigQueryRDD.scala
```diff
@@ -62,27 +58,65 @@ class BigQueryRDD(sc: SparkContext,
...
+                          rowResponseIterator : Iterator[ReadRowsResponse])
+{
+  def getIterator(): Iterator[InternalRow] =
+  {
+    // TODO: this byte string can get extremely long and may exceed memory.
+    // Figure out a way to combine iterators
+    var allRowsByteString = ByteString.EMPTY
+
+    rowResponseIterator.
+      foreach(readRowResponse => allRowsByteString = allRowsByteString.concat
```
emkornfield: this seems like it will have really poor performance.  I think you probably want to try to buffer N responses at a time and iterate over the responses with the ArrowReader.  I think pclay had a better solution to this in his PoC, did that not work?

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
emkornfield: yep, that sounds right.  could you add a comment noting BQ doesn't support this type?

connector/src/main/scala/com/google/cloud/spark/bigquery/direct/BigQueryRDD.scala
Gaurangi94: His solution had some comments with concerns regarding how he was using the buffer.
 
I understand your concern here, I have comments regarding that in lines 83 and 84. I am thinking of combining iterators wherein each iterator goes over a readRowResponse. Does that seem right?

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+    {
+      java.util.List<Field> fieldChildren = field.getChildren();
+      StructField[] structFields = new StructField[fieldChildren.size()];
+
+      int ind = 0;
+
+      for (Field childField : field.getChildren())
+      {
+        DataType childType = fromArrowField(childField);
+        structFields[ind++] = new StructField(childField.getName(), childType, childField.isNullable(), Metadata.empty());
```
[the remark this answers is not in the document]
Gaurangi94: Found from StructField documentation - "The Metadata of this field. The metadata should be preserved during transformation if the content of the column is not modified"

We do not need it in our use case and thats why it has been kept empty.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+      this.accessor = vector;
+    }
+
+    @Override
+    final long getLong(int rowId) {
+      return accessor.get(rowId);
+    }
+
+    @Override
+    final UTF8String getUTF8String(int rowId) {
```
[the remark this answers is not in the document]
Gaurangi94: Yes. To keep it consistent we are transforming this into string.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
Gaurangi94: added

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+    switch (arrowType.getTypeID())
+    {
+      case Int: return DataTypes.LongType;
+      case Bool: return DataTypes.BooleanType;
+      case FloatingPoint: return DataTypes.DoubleType;
+      case Binary: return DataTypes.BinaryType;
+      case Utf8: return DataTypes.StringType;
+      case Date: return DataTypes.DateType;
+      case Time:
+      case Timestamp: return DataTypes.TimestampType;
```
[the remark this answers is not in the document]
Gaurangi94: I am sorry I did not get the context here. Could you please elaborate?

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+
+    @Override
+    final long getLong(int rowId) {
+      return accessor.get(rowId);
+    }
+  }
+
+  private static class DoubleAccessor extends ArrowSchemaConverter.ArrowVectorAccessor {
+
+    private final Float8Vector accessor;
```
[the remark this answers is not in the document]
Gaurangi94: It seemed fine to me since it is accessing the value for corresponding row from a column vector.

connector/src/main/scala/com/google/cloud/spark/bigquery/ArrowBinaryIterator.scala
Gaurangi94: That is fine since root is just an allocator here. But thanks for pointing out! I will keep an eye on it, in case it affects anything.

connector/src/main/scala/com/google/cloud/spark/bigquery/AvroBinaryIterator.scala
Gaurangi94: I don't think so. This was part of Avro code which I refactored. I wanted to make sure to not remove any unfinished TODOs. But it does not seem like its relevant anymore.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+      this.accessor = vector;
+    }
+
+    @Override
+    final UTF8String getUTF8String(int rowId) {
+      accessor.get(rowId, stringResult);
+      if (stringResult.isSet == 0) {
+        return null;
+      } else {
+        return UTF8String.fromAddress(null,
```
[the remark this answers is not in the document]
Gaurangi94: Since it will be returned by value, does it matter? This value will be used while converting mutableColumnarRow into a Spark Row.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+
+    ArrayAccessor(ListVector vector) {
+      super(vector);
+      this.accessor = vector;
+      this.arrayData = new ArrowSchemaConverter(vector.getDataVector());
+    }
+
+    @Override
+    final boolean isNullAt(int rowId) {
+      // TODO: Workaround if vector has all non-null values, see ARROW-1948
```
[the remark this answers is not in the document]
Gaurangi94: Removed

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,497 @@
+package com.google.cloud.spark.bigquery.converters;
```
Gaurangi94: Add a license

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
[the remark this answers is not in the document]
Gaurangi94: Resolved offline.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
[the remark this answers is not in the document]
Gaurangi94: Add comment and document the issue. 

"Relying on memory buffer that may not be present. Might be better to include copy to byte array.Not incl for perf reasons"

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+      }
+    }
+
+    @Override
+    final ColumnarArray getArray(int rowId) {
+      ArrowBuf offsets = accessor.getOffsetBuffer();
+      int index = rowId * ListVector.OFFSET_WIDTH;
+      int start = offsets.getInt(index);
+      int end = offsets.getInt(index + ListVector.OFFSET_WIDTH);
+      return new ColumnarArray(arrayData, start, end - start);
```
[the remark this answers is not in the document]
Gaurangi94: code change not necessary at the moment.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
emkornfield: Update the [LICENSE file](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/master/LICENSE) as well to add specific reference to spark code.  As an example look at the [Arrow LICENSE file](https://github.com/apache/arrow/blob/master/LICENSE.txt#L210) and how it does attribution.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,492 @@
...
+    final Decimal getDecimal(int rowId, int precision, int scale) {
+      if (isNullAt(rowId)) return null;
+      return Decimal.apply(accessor.getObject(rowId), precision, scale);
+    }
+  }
+
+  private static class StringAccessor extends ArrowSchemaConverter.ArrowVectorAccessor {
+
+    private final VarCharVector accessor;
+    private final NullableVarCharHolder stringResult = new NullableVarCharHolder();
```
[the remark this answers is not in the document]
Gaurangi94: Fixed. And opened a JIRA here - https://issues.apache.org/jira/browse/ARROW-8158

Could you check if I have described the issue correctly?

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
Gaurangi94: Arrow is also under Apache License 2.0 which we already include in our project.
I am not exactly sure about what  exactly I am supposed to change here. @davidrabinowitz  could you advise accordingly?

emkornfield: I provided the Arrow file as an example of how to provide attribution to code that is copied from other projects under the apache 2.0 license file.  I think you should add an additional section to the existing license file that acknowledges code from spark was used from spark.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,521 @@
...
+  }
+
+  @Override
+  public ColumnarArray getArray(int rowId) {
+    if (isNullAt(rowId)) return null;
+    return accessor.getArray(rowId);
+  }
+
+  @Override
+  // BigQuery does not support Map type but this function needs to be overridden since this
```
emkornfield: I think inside the function would be a better place for this.  Generally you shouldn't put comments between annotations and method declaration.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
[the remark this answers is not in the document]
emkornfield: commented on the JIRA.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```diff
@@ -0,0 +1,521 @@
...
+
+    private final DateDayVector accessor;
+
+    DateAccessor(DateDayVector vector) {
+      super(vector);
+      this.accessor = vector;
+    }
+
+    @Override
+    /**
```
emkornfield: move the comment above @Override

connector/src/main/scala/com/google/cloud/spark/bigquery/ArrowBinaryIterator.scala
```diff
@@ -0,0 +1,108 @@
...
+                          schema: ByteString,
+                          rowsInBytes: ByteString) extends Iterator[InternalRow] {
+
+
+  // max allocation value for the allocator
+  var maxAllocation = Long.MaxValue
+  val allocator = ArrowUtils.rootAllocator.newChildAllocator("ArrowBinaryIterator",
+    0, maxAllocation)
+
+  val byteStringWithSchema = schema.concat(rowsInBytes)
```
emkornfield: we should avoid doing the concatenation here at all.

Use either: [SequenceInputStream](https://docs.oracle.com/javase/7/docs/api/java/io/SequenceInputStream.html) or
the pattern in in the [storage example](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/bigquery/bigquerystorage/src/main/java/com/example/bigquerystorage/StorageArrowSample.java#L83). 

 Not for this PR but it would be good to verify that creating new iterators for each rowsInBytes isn't a source of overhead.

connector/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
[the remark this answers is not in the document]
Gaurangi94: Thanks!

LICENSE
```diff
@@ -200,3 +200,14 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+--------------------------------------------------------------------------------
+
+/src/main/java/com/google/cloud/spark/bigquery/converters/ArrowSchemaConverter.java
```
Gaurangi94: @davidrabinowitz @emkornfield Could you please review this retribution?

emkornfield: Looks reasonable to me.

davidrabinowitz: I've got the response from Google legal - in your e-mail.

Gaurangi94: Updated folder structure

PR #146, opened 2020-03-24
Use access token for the Authorisation

mayurdb:
With this change, the user can pass the access token to the spark with conf `spark.gcs.user.accessToken` and the same will be used for the authorization of the big-query resources.

This change still does not support refreshing the access token on expiry which would become an issue for the long-running spark applications.

I am currently thinking of this approach to solve this - 
1. User passes the refresh token, client ID and client secret to Spark along with the access token if she wants to auto-refresh the access token.
2. When the BigQueryRelation gets created, on the driver,  generate the access token if not already available. 
3. Broadcast the generated access token so that each executor has access to it.
4. Run a daemon thread on the driver which asynchronously refreshes the access token before the expiry so that each executor always has the valid access token. ( I will have to research a bit on this part to update the value of the already broadcasted conf).

Let me know if this sounds workable or if there is some simpler approach to do this.

### DIFF
--- a/connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
+++ b/connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
@@ -90,7 +90,7 @@ object SparkBigQueryOptions {
   private val PermittedIntermediateFormats = Set(FormatOptions.orc(), FormatOptions.parquet())
   private val PermittedReadDataFormats = Set(DataFormat.ARROW.toString, DataFormat.AVRO.toString)
 
-  val GcsAccessToken = "spark.gcpAccessToken"
+  val GcsAccessToken = "gcpAccessToken"
 
   def apply(
              parameters: Map[String, String],
--- a/connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
+++ b/connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
@@ -90,7 +90,7 @@ object SparkBigQueryOptions {
   private val PermittedIntermediateFormats = Set(FormatOptions.orc(), FormatOptions.parquet())
   private val PermittedReadDataFormats = Set(DataFormat.ARROW.toString, DataFormat.AVRO.toString)
 
-  val GcsAccessToken = "spark.gcs.user.accessToken"
+  val GcsAccessToken = "spark.gcpAccessToken"
 
   def apply(
              parameters: Map[String, String],
--- a/connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
+++ b/connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
@@ -46,9 +46,9 @@ case class SparkBigQueryOptions(
     partitionType: Option[String] = None,
     createDisposition: Option[CreateDisposition] = None,
     optimizedEmptyProjection: Boolean = true,
+    accessToken: Option[String] = None,
     viewExpirationTimeInHours: Int = 24,
-    maxReadRowsRetries: Int = 3,
-    accessToken: Option[String] = None
+    maxReadRowsRetries: Int = 3
   ) {
 
   def createCredentials: Option[Credentials] =
@@ -79,7 +79,7 @@ object SparkBigQueryOptions {
   val DefaultFormat: FormatOptions = FormatOptions.parquet()
   private val PermittedIntermediateFormats = Set(FormatOptions.orc(), FormatOptions.parquet())
 
-  val GcsAccessTokenConfig = "spark.gcs.user.accessToken"
+  val GcsAccessToken = "spark.gcs.user.accessToken"
 
   def apply(
              parameters: Map[String, String],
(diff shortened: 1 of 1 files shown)


mayurdb:
@googlebot I signed it!

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -47,22 +47,24 @@ case class SparkBigQueryOptions(
     createDisposition: Option[CreateDisposition] = None,
     optimizedEmptyProjection: Boolean = true,
     viewExpirationTimeInHours: Int = 24,
-    maxReadRowsRetries: Int = 3
+    maxReadRowsRetries: Int = 3,
+    accessToken: Option[String] = None
```
davidrabinowitz: please move the `accessToken` parameter to be after `optimizedEmptyProjection` so we can keep all the constants at the end.

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -128,13 +132,14 @@ object SparkBigQueryOptions {
...
       allConf, parameters, "optimizedEmptyProjection", true)
+    val accessToken = getAnyOption(allConf, parameters, GcsAccessTokenConfig)
 
     SparkBigQueryOptions(tableId, parentProject, credsParam, credsFileParam,
       filter, schema, maxParallelism, temporaryGcsBucket, intermediateFormat,
       combinePushedDownFilters, viewsEnabled, materializationProject,
       materializationDataset, partitionField, partitionExpirationMs,
       partitionRequireFilter, partitionType, createDisposition,
-      optimizedEmptyProjection)
+      optimizedEmptyProjection, accessToken = accessToken)
```
davidrabinowitz: Once the parameter has moved, please remove the `= ...` part

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -77,6 +79,8 @@ object SparkBigQueryOptions {
   val DefaultFormat: FormatOptions = FormatOptions.parquet()
   private val PermittedIntermediateFormats = Set(FormatOptions.orc(), FormatOptions.parquet())
 
+  val GcsAccessTokenConfig = "spark.gcs.user.accessToken"
```
davidrabinowitz: Please change the name to `gcpAccessToken`

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
mayurdb: Done

mayurdb:
@davidrabinowitz can you please also reply to the comment above?

davidrabinowitz:
I was hoping to get the initial AccessToken support ready as  we plan to release the next version of the connector next week (March 30/31).

In the long run, something along the points you have outlined sounds like a good plan. Some of the work I've started for implemented DataSource v2 will probably make it easier. I suggest that we merge this PR as is (just changing the option name), and let's open another issue where we can continue the discussion.

apostaremczak:
Hey, are you planning on merging this feature and including in the the next release? I'd like to use it as well :)

mayurdb:
@davidrabinowitz  I have addressed the comments and tested the changes on my end -
`
//set access token for the user who has read permission to the dataset
> spark.sql("SET spark.gcs.user.accessToken=<access_token>").collect

// Verify the user has the access
> val df = spark.read.format("bigquery").option("table","mayur.customTable").load()
> df.collect()
> res6: Array[org.apache.spark.sql.Row] = Array([75,75,key75,dataPoint75], [76,76,key76,dataPoint76], [77,77,key77,dataPoint77], ...

//set access token for the user who does not have read permission to the dataset
> spark.sql("SET spark.gcs.user.accessToken=<no_access_user_access_token>").collect

// Verify that this user does not have the access
> val df = spark.read.format("bigquery").option("table","mayur.customTable").load()
> df.collect()
com.google.cloud.spark.bigquery.repackaged.com.google.cloud.bigquery.BigQueryException: Access Denied: Table spark-cutomer-project-1:mayur.partitionedTable: User does not have bigquery.tables.get permission for table spark-cutomer-project-1:mayur.partitionedTable.

// Give this user access via the GCP console and verify that the user is able to read the dataset
> val df = spark.read.format("bigquery").option("table","mayur.customTable").load()
> df.collect()
res27: Array[org.apache.spark.sql.Row] = Array([75,75,key75,dataPoint75], [76,76,key76,dataPoint76], [77,77,key77,dataPoint77], [78,78,key78,dataPoint78],....
`

I have verified this for the writes as well. Anything else pending from my side?

davidrabinowitz:
Hi @mayurdb 

Can you please do the follwoing:

- Replace the parameter name from `spark.gcs.user.accessToken` to `gcpAccessToken` in order to keep name consistency? Also the token is for all GCP authentication.
- Rebase the PR to master? Currently I cannot merge it due to conflicts.

I plan to release a new version next week, and I'd be happy to include this PR in it.

mayurdb:
@davidrabinowitz  I have changed the conf name to `spark.gcpAccessToken` which is the spark convention of naming confs. Let me know if you had like me to change this to just gcpAccessToken :)

I have also rebased the branch to master!

davidrabinowitz:
@mayurdb Yes, I'd appreciate it if you can change it to just `gcpAccessToken` in order to keep consistency with the rest of the parameters. I know that the spark convention is to stark with `spark.`, but at the moment backwards compatibility is important and I don't want to break existing applications relying on the connector.

mayurdb:
@davidrabinowitz  done

mayurdb:
@davidrabinowitz  just realized, spark ignores the confs which are not prefixed with `spark.*` So the users might not be able to pass just `gcpAccessToken`

PR #147, opened 2020-03-27
Exposing ServiceUpdateOption to be added when writing to BigQuery

davidrabinowitz:


README.md
```diff
@@ -310,6 +310,24 @@ The API Supports a number of options to configure the read
         </td>
         <td>Write</td>
      </tr>
+   <tr valign="top">
```
Gaurangi94: Shouldn't we be documenting "loadSchemaUpdateOption" here since that is how we are exposing this feature to the users?

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -143,12 +148,24 @@ object SparkBigQueryOptions {
     val optimizedEmptyProjection = getAnyBooleanOption(
       allConf, parameters, "optimizedEmptyProjection", true)
 
+    val allowFieldAddition = getAnyBooleanOption(
```
Gaurangi94: do we need unit tests for this? Just to make sure that the option value got updated correctly.

README.md
davidrabinowitz: We are not exposing `loadSchemaUpdateOption`, it's an internal representation. We are exposing the `allowFieldAddition` and `allowFieldRelaxation` options

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
davidrabinowitz: Of course, pushed them now.

PR #150, opened 2020-04-07
perf: Using an iterator instead of going through all the data

giomerlin:
@davidrabinowitz Comparing the version 0.9.2-beta with the current one I realized that it was pretty slow when doing limit or show operations. It was due the introduction of the ReadRowsHelper in the version 0.10.0-beta which was iterating over all the results before returning the final iterator, so I wrapped the results in a new iterator, achieving the same retry functionality


davidrabinowitz:
Thanks @giomerlin for this PR! This task was on my to-do list for 0.15... Can you please revert the changes to build.sbt?

giomerlin:
@davidrabinowitz I didn't realize I pushed the sbt changes, I've just reverted them

connector/src/main/scala/com/google/cloud/spark/bigquery/direct/BigQueryRDD.scala
```diff
@@ -139,41 +139,54 @@ case class ReadRowsClientWrapper(client: BigQueryStorageClient)
...
-            client.close
+            helper.client.close
             throw e
           }
       }
-    }
-    readRowResponses.iterator
+    } while (serverResponses.hasNext)
+
+    null
```
davidrabinowitz, line 173: Per `java.util.Iterator.next()` [javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#next--), a `NoSuchElementException` should be thrown

giomerlin: Done :)

davidrabinowitz:
Have you made changes to `ReadRowsSuite.scala`? Unfortunately it fails now.

giomerlin:
@davidrabinowitz I just pushed the changes to the test, forgot to run the test after I made the scope change to the fetchResponses method, It should run now

PR #160, opened 2020-04-23
Do filter push down when reading from views

diggerk:
Here's a fix for filter push down for views. I've moved temp table creation to the moment when "buildScan" is invoked. The defaultTableDefinition type is changed to "TableDefinition" and a separate method "getNumBytes" is created to calculate the size of the table. The "getNumBytes" returns zero bytes for views, and that should be fine as the important part is that the connector calculates the number of partitions after the temp table is generated and by that time the getNumBytes will be called on the temp table, not the view.

### DIFF
--- a/connector/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
+++ b/connector/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
@@ -70,10 +70,9 @@ private[bigquery] class DirectBigQueryRelation(
   val DEFAULT_BYTES_PER_PARTITION = 400L * 1000 * 1000
 
   override val needConversion: Boolean = false
-  override val sizeInBytes: Long = defaultTableDefinition.getNumBytes
+  override val sizeInBytes: Long = getNumBytes(defaultTableDefinition)
   // no added filters and with all column
-  lazy val defaultTableDefinition: StandardTableDefinition =
-    getActualTable(Array(), Array()).getDefinition[StandardTableDefinition]
+  lazy val defaultTableDefinition: TableDefinition = table.getDefinition[TableDefinition]
 
   override def buildScan(): RDD[Row] = {
     buildScan(schema.fieldNames)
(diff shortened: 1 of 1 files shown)


connector/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
```diff
@@ -269,10 +267,19 @@ private[bigquery] class DirectBigQueryRelation(
...
+  def getMaxNumPartitionsRequested(tableDefinition: TableDefinition): Int =
     options.maxParallelism
       .getOrElse(Math.max(
-        (tableDefinition.getNumBytes / DEFAULT_BYTES_PER_PARTITION).toInt, 1))
+        (getNumBytes(tableDefinition) / DEFAULT_BYTES_PER_PARTITION).toInt, 1))
+
+  def getNumBytes(tableDefinition: TableDefinition): Long = {
+    val tableType = tableDefinition.getType
+    if (options.viewsEnabled && TableDefinition.Type.VIEW == tableType) {
+      0
```
davidrabinowitz, line 278: Please replace the value with the default: `sqlContext.conf.defaultSizeInBytes` (taken from [here](https://github.com/apache/spark/blob/v2.4.0/sql/core/src/main/scala/org/apache/spark/sql/sources/interfaces.scala#L212))
Based on the documentation having a size of 0 may have side effects.

davidrabinowitz:
Hi @diggerk , thanks for the contribution! Can you please have a look at the comment above?

connector/src/main/scala/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.scala
digger: `sqlContext.conf` is not accessible outside of `org.apache.spark.sql` so I had to use `sqlContext.sparkSession.sessionState.conf.defaultSizeInBytes` instead

digger:
Hi @davidrabinowitz , thanks for the catch, I fixed that.

PR #162, opened 2020-04-29
Add Avro as intermediate format

Gaurangi94:


Gaurangi94:
@googlebot I fixed it.

PR #163, opened 2020-04-29
Add avro as intermediate format

Gaurangi94:


connector/src/test/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptionsSuite.scala
```diff
@@ -35,6 +36,8 @@ class SparkBigQueryOptionsSuite extends FunSuite {
         parameters,
         Map.empty[String, String], // allConf
         hadoopConfiguration,
+        new SQLConf,
+        "2.4.0",
```
davidrabinowitz, line 40: Please convert to a constant (`SparkVersion`) rather than use a magic String. It applies to the rest of the uses of this String in the other tests

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -187,6 +204,24 @@ object SparkBigQueryOptions {
       optimizedEmptyProjection, accessToken, loadSchemaUpdateOptions.asJava)
   }
 
+  // could not load the spark-avro data source
+  private def missingAvroException(sparkVersion: String, cause: Exception) = {
+    val message = if (sparkVersion >= "2.4") {
```
davidrabinowitz, line 209: Please make the error message always display the correct avro package:
```
val package = if (sparkVersion >= "2.4") {
  val scalaVersion = Properties.versionNumberString
  val scalaShortVersion = scalaVersion.substring(0, scalaVersion.lastIndexOf('.'))
  s"org.apache.spark:spark-avro_$scalaShortVersion:$sparkVersion"
} else {
  "com.databricks:spark-avro_2.11:4.0.0"
}
val message = s"""Avro writing is not supported, as the spark-avro has not been
  |found. Please re-run spark with the --packages $package parameter"""
  .stripMargin.replace('\n', ' ')
```

PR #166, opened 2020-04-30
Map Avro intermediate format

Gaurangi94:


connector/src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndITSuite.scala
```diff
@@ -474,6 +480,18 @@ class SparkBigQueryEndToEndITSuite extends FunSuite
     assert(df.schema == allTypesTable.schema)
   }
 
+  /* test("write all types to bq - json format") {
+    writeToBigQuery(allTypesTable, SaveMode.Overwrite, "json")
```
Gaurangi94, line 484: @davidrabinowitz  Currently this test fails because the ordering of data read from big query in json format is different from how it was loaded. We can discuss this further

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -270,6 +242,66 @@ object SparkBigQueryOptions {
...
+
+    def apply (format: String, sparkVersion: String, sqlConf: SQLConf) : IntermediateFormat = {
+      if (!PermittedIntermediateFormats.contains(format)) {
+        throw new IllegalArgumentException(
+          s"""Data read format '${format}' is not supported.
+             |Supported formats are '${PermittedIntermediateFormats.keys.mkString(",")}'"""
+            .stripMargin.replace('\n', ' '))
+      }
+
+      val formatSuffix = if (format == "json") {
```
davidrabinowitz, line 264: What's the difference between json and parquet in this case? why not to have `formatSuffix = s".$format"`?

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -270,6 +242,66 @@ object SparkBigQueryOptions {
...
+        val dataSource = if (sparkVersion >= "2.4") {
+          format
+        } else {
+          "com.databricks.spark.avro"
+        }
+
+        try {
+          DataSource.lookupDataSource(dataSource, sqlConf)
+        } catch {
+          case re: org.apache.spark.sql.AnalysisException =>
```
davidrabinowitz, line 280: Add import for `AnalysisException`, rename to ae (re was for `RuntimeException`)

connector/src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndITSuite.scala
davidrabinowitz: That's why we need to add the schema as well.

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
Gaurangi94: We had formatoption.type.tolowercase as the suffix earlier. This does not work for json since type gets resolve to newline_delimited_json.

davidrabinowitz: I thought it was the case when the value was of type `FormatOptions`. Now, if I'm not mistaken it is one of the Strings ("avro", "json", "parquet", "orc"), so we can use it directly

Gaurangi94: okay great. I thought there was a techincal reason for getting the type from FormatOptions

PR #171, opened 2020-05-26
Adding tests and format fix

davidrabinowitz:


connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryStorageClientFactory.java
```diff
@@ -1,3 +1,18 @@
+/*
```
Gaurangi94, line 1: Seems like a copyright statement is already there below this one.

PR #176, opened 2020-05-29
Simplifying the usage

davidrabinowitz:
Simplifying the usage of the connector, instead of the `table` option, the usage can be the more common:
```
df = spark.read.format("bigquery").load(<table>)
df.write.format("bigquery").save(<table>)
```


davidrabinowitz:
Simplifying the usage of the connector, instead of the table option, the usage can be the more common:

df = spark.read.format("bigquery").load(<table>)
df.write.format("bigquery").save(<table>)

PR #198, opened 2020-06-27
Adds implementation for supporting columnar batch reads from Spark.

emkornfield:
This bypasses most of the existing translation code for the following reasons:
1.  I think there might be a memory leak because the existing code doesn't close the allocator.
2.  This avoids continuously recopying the schema.

I didn't delete the old code because it appears the BigQueryRDD still relies on it partially.

I also couldn't find instructions on formatting/testing (I couldn't find explicit unit tests
for existing arrow code, I'll update accordingly if pointers can be provided).

### DIFF
--- a/connector/src/main/java/com/google/cloud/spark/bigquery/v2/ArrowInputPartition.java
+++ b/connector/src/main/java/com/google/cloud/spark/bigquery/v2/ArrowInputPartition.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


emkornfield:
CC @Gaurangi94 @davidrabinowitz

emkornfield:
looks like I don't have permissions to see the build failure?

davidrabinowitz:
Failed on a timeout, re-ran

emkornfield:
Looks like with a fix to helper class it passes gcbrun?  I had to force push do to the change that reformatted the entire code base.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/ArrowInputPartition.java
```diff
@@ -0,0 +1,47 @@
+package com.google.cloud.spark.bigquery.v2;
```
emkornfield, line 1: should this have a license header?

davidrabinowitz: That's what we need

Issue #200, opened 2020-07-01
(Py)Spark 3.0 / Java 11 fails with `java.lang.UnsupportedOperationException: sun.misc.Unsafe or java.nio.DirectByteBuffer.<init>(long, int) not available`


jonesberg:
When using the connector with Spark 3.0.0 (on Java 11), trying to read a dataset from BigQuery fails with the error at the bottom.

This is a known problem with Java 9+ and Spark as shown on their documentation (https://spark.apache.org/docs/3.0.0/). Here is the pull request in question 
https://github.com/apache/spark/pull/26552

> For Java 11, `-Dio.netty.tryReflectionSetAccessible=true` is required additionally for Apache Arrow library. This prevents `java.lang.UnsupportedOperationException: sun.misc.Unsafe or java.nio.DirectByteBuffer.(long, int)` not available when Apache Arrow uses Netty internally.

My Spark instance is launched with the `-Dio.netty.tryReflectionSetAccessible=true` flags enabled and Pandas UDF/Arrow conversion are working. I downloaded a sample data set from BigQuery to test my code and it works without any issues.

**Steps to reproduce**

With Spark 3.0.0 / Java 11.
Launch PySpark with the `--conf spark.driver.extraJavaOptions="-Dio.netty.tryReflectionSetAccessible=true" --conf spark.executor.extraJavaOptions="-Dio.netty.tryReflectionSetAccessible=true"` options.

Try to read a BigQuery dataset

```python
df =   (spark.read.format("bigquery")
        .option("table", f"bigquery-public-data.noaa_gsod.gsod1929")
        .option("credentialsFile", "bq-key.json")
        .load())

df.show()
```

Stacktrace

```
20/07/01 11:53:14 ERROR Executor: Exception in task 3.0 in stage 0.0 (TID 3)
java.lang.UnsupportedOperationException: sun.misc.Unsafe or java.nio.DirectByteBuffer.<init>(long, int) not available
	at com.google.cloud.spark.bigquery.repackaged.io.netty.util.internal.PlatformDependent.directBuffer(PlatformDependent.java:399)
	at com.google.cloud.spark.bigquery.repackaged.io.netty.buffer.NettyArrowBuf.getDirectBuffer(NettyArrowBuf.java:257)
	at com.google.cloud.spark.bigquery.repackaged.io.netty.buffer.NettyArrowBuf.nioBuffer(NettyArrowBuf.java:247)
[...]

PR #204, opened 2020-07-06
Upgrade Netty to a version supporting Java 9+ (Issue #200)

davidrabinowitz:


build.sbt
```diff
@@ -16,6 +16,7 @@
 lazy val scala211Version = "2.11.12"
 lazy val scala212Version = "2.12.10"
 lazy val sparkVersion = "2.4.0"
+lazy val nettyVersion = "4.1.30.Final"
```
medb, line 19: Should we use the latest Netty `4.1.50.Final` version: https://search.maven.org/artifact/io.netty/netty-all ?

Also, you should keep in sync Netty and gRPC versions: https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty

Or you can avoid this sync issue by using `grpc-netty-shaded` gRPC artifact instead of Netty artifacts.

davidrabinowitz: We are using the shaded version. I've upgraded to `4.1.50.Final`

medb: In this case you should drop all `io.netty:` dependencies because you are using `grpc-netty-shaded` dependency that has an appropriate version for the gRPC version.

davidrabinowitz: I can't - Arrow adds dependency on netty in version `4.1.27.Final` so we need to explicitly upgrade those.

medb: Interesting, in this case we should not use `grpc-netty-shaded` for gRPC, because BQ connector will have 2 Netty artifacts - one shaded for gRPC and one not shaded for Arrow? In this case, I think, both gRPC and Arrow should use the same unshaded Netty.

charlesa101:
Looks this error still exist in 

openjdk 11.0.10 2021-01-19
OpenJDK Runtime Environment 18.9 (build 11.0.10+9)

Issue #196, opened 2020-06-25
Bug when reading views


afghifariii:
when reading to a `view` on a different project, it always try to create the job in the default `projectId`. here is the example
for context i am running the spark job also on different project let say `current-project-id`

```
val parentProjectId = "parent-project-id"
val anotherProjectid = "another-project-id"
val fullTableName = s"$anotherProjectId:some_table.some_view"
val viewMaterializationDataset = "some_tmp_dataset"

val df = spark.read
    .option("parentProject", parentProjectId)
    .option("viewsEnabled", true)
    .option("viewMaterializationProject", parentProjectId)
    .option("viewMaterializationDataset", viewMaterializationDataset)
    .bigquery(fullTableName)
```

from my understanding, it should create a `materialized view` on project `parent-project-id` and dataset `some_tmp_dataset`, but what happened is it try to create the `materialized view` using `current-project-id` because i get permission issue because the ServiceAccount doesn't have permission `job.create` in `current-project-id`

davidrabinowitz:
Can you please set the log to debug and look for the log line that begins with `destinationTable is`? This should give us the materialization table.

afghifariii:
```
main DEBUG DirectBigQueryRelation: destinationTable is GenericData{classInfo=[datasetId, projectId, tableId], {datasetId=some_tmp_dataset, projectId=parent-project-id, tableId=_sbc_e696652833fa4c4cb914b663e9e89eef}}
```

kmjung:
At a high level, from BigQuery's perspective, it's possible for all of these projects to be different:
- The project which owns the view being read from
- The project which owns the job to query the view and is billed for the query execution
- The project which owns the destination table to which data is written by the query
- The project which reads the data (e.g. the parent project for the read session) and is billed for the read

It's also possible for the project in which the Dataproc job is running to be distinct from all of the above.

I'm not familiar with what the connector actually does here, but my expectation would be that, by default, if you've specified a `"parentProject"` option, then the query job to read from the view should be created in the context of that project, and that the output table should also be created in the context of that project. It sounds like this is not the case today, and that the job to query the view is actually created in the context of the "current" project (e.g. the one associated with the service account in which your job is running). @davidrabinowitz does that sound right to you?

Note that [BigQuery materialized views](https://cloud.google.com/bigquery/docs/materialized-views-intro) are not involved here -- the connector simply executes a `SELECT *` query against the target view and writes the result to a simple destination table.

[...]

PR #208, opened 2020-07-10
set projectId when create BigQueryClient (Issue #196)

afghifariii:


afghifariii:
i was wrong, turns out it didnt' affect to where the `BigQuery Jobs` created

PR #210, opened 2020-07-10
Performance tuning of Arrow

Gaurangi94:


connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -146,6 +148,23 @@ object SparkBigQueryOptions {
           .stripMargin.replace('\n', ' '))
     }
     val readDataFormat = DataFormat.valueOf(readDataFormatParam)
+
+    val arrowEnableUnsafeMemoryAccessProperty = "arrow.enable_unsafe_memory_access"
+    val arrowEnableUnsafeMemoryAccess = getAnyBooleanOption(normalizedAllConf,
+      parameters,
+      arrowEnableUnsafeMemoryAccessProperty,
+      DefaultArrowUnsafeMemoryAccess)
+    System.setProperty(arrowEnableUnsafeMemoryAccessProperty,
```
davidrabinowitz, line 157: This will be set on the driver process only, not on the executors. The property needs to be set in the ArrowBinaryIterator/Reader as well (and same for the other property)

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
```diff
@@ -146,6 +148,23 @@ object SparkBigQueryOptions {
           .stripMargin.replace('\n', ' '))
     }
     val readDataFormat = DataFormat.valueOf(readDataFormatParam)
+
+    val arrowEnableUnsafeMemoryAccessProperty = "arrow.enable_unsafe_memory_access"
```
davidrabinowitz, line 152: Until we unify SpringBigQueryOptions and SparkBIgQueryConfig, please implement the setting in both places, so that the DataSource v2 implementation will have this as well

Gaurangi94: Added corresponding changes to SparkBigQueryConfig and SparkBigQueryConfigSuite. Is that enough?

connector/src/main/scala/com/google/cloud/spark/bigquery/SparkBigQueryOptions.scala
Gaurangi94: Added it in BigQueryRDD as well

davidrabinowitz:
Unfortunately this can be solved by documentation change only, as arrow reads those parameters on load.

PR #212, opened 2020-07-14
Update streaming write doc

varundhussa:
Updating readme to include streaming writes. 
Minor typo fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,17 @@ df.write
   .save("dataset.table")
 ```
 
-*Inportant:* The connector does not configure the GCS connector, in order to avoid conflict with another GCS connector, if exists. In order to use the write capabilities of the connector, please configure the GCS connector on your cluster as explained [here](https://github.com/GoogleCloudPlatform/bigdata-interop/tree/master/gcs).
+When writing a streaming DataFrame to BigQuery, each batch is written in the same manner as a non-streaming DataFrame. Note that a HDFS compatible checkpoint location must be specified.
+
+```
+df.writeStream
+  .format("bigquery")
+  .option("temporaryGcsBucket","some-bucket")
+  .option("checkpointLocation", "some-location")
+  .option("table", "dataset.table")
+```
+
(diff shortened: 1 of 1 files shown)


varundhussa:
@davidrabinowitz Added streaming writes to the readme

README.md
```diff
@@ -136,7 +136,17 @@ df.write
...
 ```
 
-*Inportant:* The connector does not configure the GCS connector, in order to avoid conflict with another GCS connector, if exists. In order to use the write capabilities of the connector, please configure the GCS connector on your cluster as explained [here](https://github.com/GoogleCloudPlatform/bigdata-interop/tree/master/gcs).
+When writing a streaming DataFrame to BigQuery, each batch is written in the same manner as a non-streaming DataFrame. Note that a HDFS compatible checkpoint location must be specified.
+
+```
+df.writeStream
+  .format("bigquery")
+  .option("temporaryGcsBucket","some-bucket")
+  .option("checkpointLocation", "some-location")
```
davidrabinowitz, line 145: When reading this, "Location" is not clear. Can we change it to "Path" ?

README.md
```diff
@@ -136,7 +136,17 @@ df.write
   .save("dataset.table")
 ```
 
-*Inportant:* The connector does not configure the GCS connector, in order to avoid conflict with another GCS connector, if exists. In order to use the write capabilities of the connector, please configure the GCS connector on your cluster as explained [here](https://github.com/GoogleCloudPlatform/bigdata-interop/tree/master/gcs).
+When writing a streaming DataFrame to BigQuery, each batch is written in the same manner as a non-streaming DataFrame. Note that a HDFS compatible checkpoint location must be specified.
```
davidrabinowitz, line 139: Please give an example for "HDFS compatible checkpoint location"

README.md
varundhussa: This is a Spark streaming sink option, so can't change it. Have updated the doc to point to the spark documentation

README.md
varundhussa: Done

PR #213, opened 2020-07-15
Spark streaming real time

varundhussa:
Use BigQuery insertAll API to stream a DataFrame
Retry and exponential backoff while calling the API


davidrabinowitz:
Closing this PR as we are going to use the BigQuery Storage Write API for streaming as well

PR #245, opened 2020-09-24
Using spark default parallelism as default

davidrabinowitz:


emkornfield (approved):
This seems reasonable to me.  One thing that occurred to me is if we want to still use default parallelism if the table size is below the broadcast threshold?

davidrabinowitz:
What happens in auto-scaling?

davidrabinowitz:
Discarded in favor of PR #668

Issue #249, opened 2020-10-02
Unable to authenticate using 'credentials' option


sercanersoy:
Hi,

I am writing a Spark (2.4.3) application using Scala (2.11.12) which reads data from a BigQuery table with the help of `spark-bigquery-with-dependencies` package (0.17.2). The application does not run on the Google Cloud machines, therefore it has to authenticate. Below is my code:

```
spark.read
  .option("parentProject", "xxx")
  .option("credentials", "xxx")
  .bigquery("xxx")
  .limit(10)
  .show()
```

I mustn't use `credentialsFile` option. Therefore, I converted my json credentials file to a Base64 string and I am passing it to `credentials` option. But this is what I got:

```
java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

The program is unable to infer credentials from `credentials` option. Actually I think it does not even trying to do so, it just directly looks at the `GOOGLE_APPLICATION_CREDENTIALS` and crashes if it does not exist. I mustn't use an external json credential file in my app hence I mustn't use an environment variable, too.

I would be appreciated if you could help. Thanks by now!

davidrabinowitz:
Can you please try the following:
```
spark.conf.set("credentials", "<SERVICE_ACCOUNT_JSON_IN_BASE64>")
spark.read
  .option("parentProject", "xxx")
  .bigquery("xxx")
  .limit(10)
  .show()
```

davidrabinowitz:
Fixed by PR #250

gbougeard:
Hi,

not sure it's the best place to post that but I have the following error using 0.17.3 
```
java.io.IOException: Error getting access token from metadata server at: http://metadata/computeMetadata/v1/instance/service-accounts/default/token
[info]   at com.google.cloud.hadoop.util.CredentialFactory.getCredentialFromMetadataServiceAccount(CredentialFactory.java:208)
[info]   at com.google.cloud.hadoop.util.CredentialConfiguration.getCredential(CredentialConfiguration.java:70)
```

I'm doing my setup like that and my spark job will run on AWS EMR: 

```
 sparkSession.conf.set("credentials", base64credentials)
 sparkSession.conf.set("parentProject", projectId)      
[...]

PR #250, opened 2020-10-05
Code simplification to solve Issue #249

davidrabinowitz:
It appears that the previous code that had relied on Java streams had issues, reverting to simpler code structure, added additional tests.

### DIFF
--- a/connector/src/test/java/com/google/cloud/bigquery/connector/common/BigQueryCredentialsSupplierTest.java
+++ b/connector/src/test/java/com/google/cloud/bigquery/connector/common/BigQueryCredentialsSupplierTest.java
@@ -1,0 +1,98 @@
+package com.google.cloud.bigquery.connector.common;
+
+import com.google.api.client.json.GenericJson;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.auth.Credentials;
+import com.google.auth.oauth2.ServiceAccountCredentials;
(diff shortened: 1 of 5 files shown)


connector/src/test/java/com/google/cloud/bigquery/connector/common/BigQueryCredentialsSupplierTest.java
```diff
@@ -0,0 +1,98 @@
...
+      "36680232662-vrd7ji19qe3nelgchd0ah2csanun6bnr@developer.gserviceaccount.com";
+  private static final String CLIENT_ID =
+      "36680232662-vrd7ji19qe3nelgchd0ah2csanun6bnr.apps.googleusercontent.com";
+  private static final String PRIVATE_KEY_ID = "d84a4fefcf50791d4a90f2d7af17469d6282df9d";
+  private static final String ACCESS_TOKEN = "1/MkSJoj1xsli0AccessToken_NKPY2";
+  private static final Collection<String> SCOPES = Collections.singletonList("dummy.scope");
+  private static final String QUOTA_PROJECT = "sample-quota-project-id";
+
+  @Test
+  public void testCredentialsFromKey() throws Exception {
```
Gaurangi94, line 50: should we also have tests for cases where we use default credentials or credentials from access token?

Gaurangi94 (approved):

PR #262, opened 2020-11-03
Convert filters on BigQuery dates and timestamp fields using the right type

anoopj:
This fixes the query failures similar to TPCDS query# 83. Without this fix, BigQuery read API complains that the pushed down filter is invalid. The fix is done for both the v1 and v2 code paths.

Testing: unit testing and tested with TPCDS query# 83

### DIFF
--- a/connector/src/main/java/com/google/cloud/spark/bigquery/SparkFilterUtils.java
+++ b/connector/src/main/java/com/google/cloud/spark/bigquery/SparkFilterUtils.java
@@ -268,8 +268,11 @@ public class SparkFilterUtils {
     if (value instanceof String) {
       return "'" + escape((String) value) + "'";
     }
-    if (value instanceof Timestamp || value instanceof Date) {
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -24,7 +24,7 @@ lazy val nettyTcnativeVersion = "2.0.29.Final"
 
 lazy val commonSettings = Seq(
   organization := "com.google.cloud.spark",
-  version := "0.17.4-SNAPSHOT",
+  version := "0.17.5-SNAPSHOT",
```
davidrabinowitz, line 27: Please remove, the next version is either 0.17.4 or 0.18.0

davidrabinowitz (commented):
/gcbrun

build.sbt
anoopj: Done.

davidrabinowitz (approved):
Thanks, there one minor fix

PR #263, opened 2020-11-05
Handle select * case in columnar batch read from Spark

mingjialiu:
This PR fixes the failure that happens with select * queries when columnar batch read is enabled. Without this fix,  TPC-DS queries including 72, 21,.. will fail. 

This can be reproed with  SELECT *  and spark-bigquery-with-dependencies jar released after PR : https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/237  (which enabled arrow format in V2)

Test : added 1 integration test case


davidrabinowitz:
Thanks! Can you please add an integration test testing this?

davidrabinowitz (approved):

PR #266, opened 2020-11-07
Fixing external configuration regression bug

davidrabinowitz:
A regression bug preventing the usage of the `--conf` parameter to configure the connector's option has been introduced in version 0.17.2. This PR fixes that


mingjialiu (approved):
This change normalized spark configs that start with spark.datasource.bigquery. So they can be retrieved at SparkBigQueryConfig initialization

mingjialiu (approved):

animeshnandanwar (approved):

PR #277, opened 2020-11-23
Estimating DataFrame statistics based on the BigQuery data

davidrabinowitz:


connector/src/test/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatisticsTest.java
```diff
@@ -0,0 +1,122 @@
+/*
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
medb, line 2: 2020 Google LLC

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
medb, line 2: 2020 Google LLC

medb (changes requested):

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -168,32 +171,57 @@ String fullTableName(TableId tableId) {
...
+      OptionalLong numberOfFilteredRows =
+          toOptionalLong(
+              filter
+                  .map(f -> countWithFilter(tableInfo.getTableId(), f))
+                  .map(row -> row.get(0).getLongValue()));
+      return new TableStatistics(tableInfo.getNumRows().longValue(), numberOfFilteredRows);
+    } else if (type == TableDefinition.Type.VIEW
+        || type == TableDefinition.Type.MATERIALIZED_VIEW) {
+      // For views, we need to calculate the number of rows
+      String table = fullTableName(tableInfo.getTableId());
```
emkornfield, line 191: I think this should be user configurable, to either return null statistics OR execute this logic.  By default I don't think we should be running these queries unless the user opts into them.  It essentially amounts to 2 data scans if the filters are completely against metadata.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -168,32 +171,57 @@ String fullTableName(TableId tableId) {
...
+      String sql =
+          filter
+              .map(
+                  f ->
+                      format(
+                          "SELECT (SELECT COUNT(*) from `%s`), (SELECT COUNT(*) from `%s` WHERE %s)",
```
emkornfield, line 197: did you compare the plan for this with something like "Select count(*), sum(filter) from table?

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -168,32 +171,57 @@ String fullTableName(TableId tableId) {
...
-                "Unsupported table type %s for table %s",
-                type, fullTableName(tableInfo.getTableId())));
-      }
-    } catch (InterruptedException e) {
-      throw new BigQueryConnectorException(
-          "Querying table size was interrupted on the client side", e);
+  public TableStatistics calculateTableSize(TableInfo tableInfo, Optional<String> filter) {
+    TableDefinition.Type type = tableInfo.getDefinition().getType();
+    if (type == TableDefinition.Type.TABLE) {
+      // if this is a table, there's no need to query the total number of rows
```
emkornfield, line 181: try to be consistent with comment grammar (capitalized below and now here).  Also, mention why a query isn't required?

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -168,32 +171,57 @@ String fullTableName(TableId tableId) {
...
   }
+
+  private OptionalLong toOptionalLong(Optional<Long> opt) {
+    return opt.map(LongStream::of).orElseGet(LongStream::empty).findFirst();
+  }
+
+  private FieldValueList countWithFilter(TableId tableId, String filter) {
+    String table = fullTableName(tableId);
+    String sql = format("SELECT COUNT(*) from `%s` WHERE %s", table, filter);
+    return queryForSingleRow(sql);
```
emkornfield, line 220: again, I think this should be opt in.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceReader.java
```diff
@@ -249,27 +243,12 @@ public void pruneColumns(StructType requiredSchema) {
...
-
-  @Override
-  public OptionalLong sizeInBytes() {
-    return OptionalLong.of(tableDefinition.getNumBytes());
-  }
-
-  @Override
-  public OptionalLong numRows() {
-    return OptionalLong.of(tableDefinition.getNumRows());
+    logger.debug("estimateStatistics()");
```
emkornfield, line 246: maybe include more info here, or remove?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
```diff
@@ -0,0 +1,107 @@
...
+import com.google.cloud.bigquery.TableInfo;
+import com.google.cloud.bigquery.connector.common.BigQueryClient;
+import com.google.cloud.bigquery.connector.common.TableStatistics;
+import com.google.common.collect.ImmutableList;
+import org.apache.spark.sql.sources.v2.reader.Statistics;
+
+import java.util.Optional;
+import java.util.OptionalLong;
+
+class BigQueryEstimatedTableStatistics implements Statistics {
```
emkornfield, line 30: docs?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
```diff
@@ -0,0 +1,107 @@
...
+      long numberOfRows = tableStatistics.getNumberOfRows();
+      long numberOfQueriedRows = tableStatistics.getQueriedNumberOfRows();
+      // Step 3 - if there where selected columns, estimate the size of the table for the entire
+      // table. Otherwise, take it from the tableInfo.
+      long projectedTableSizeInBytes =
+          selectedColumns
+              .map(this::getProjectedTableSizeInBytesFromColumns)
+              .orElse(table.getNumBytes());
+      long sizeInBytes =
+          (long) (1.0 * projectedTableSizeInBytes * numberOfQueriedRows / numberOfRows);
```
emkornfield, line 92: can numberOfRows be zero?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
```diff
@@ -0,0 +1,107 @@
...
+      this.filter = filter;
+      return this;
+    }
+
+    Builder setSelectedColumns(Optional<ImmutableList<String>> selectedColumns) {
+      this.selectedColumns = selectedColumns;
+      return this;
+    }
+
+    BigQueryEstimatedTableStatistics build() {
```
emkornfield, line 78: build is slightly misleading (at least for me).  I expect builders to not really have external logic.  I don't have a better suggestion at the moment through.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
```diff
@@ -0,0 +1,107 @@
...
+      // step 1 - get the statistics
+      TableStatistics tableStatistics = bigQueryClient.calculateTableSize(table, filter);
+      // step 2 - if there is a filter, then issue a query to find the number of rows by the query.
+      // Otherwise, take it from the tableInfo.
+      long numberOfRows = tableStatistics.getNumberOfRows();
```
emkornfield, line 83: is it possible to move this method to TableStatistics or get rid of TableStatistics?

connector/src/main/java/com/google/cloud/bigquery/connector/common/TableStatistics.java
```diff
@@ -0,0 +1,28 @@
+package com.google.cloud.bigquery.connector.common;
+
+import java.util.Optional;
+import java.util.OptionalLong;
+
+public class TableStatistics {
```
emkornfield, line 6: docs.  Also it seems that there is TableStatistics and EstimatedTableStatistics, is there some way to change the naming to differentiate these a bit more?

connector/src/test/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatisticsTest.java
```diff
@@ -0,0 +1,124 @@
...
+    when(TABLE.getDefinition()).thenReturn(TABLE_DEFINITION);
+  }
+
+  @Test
+  public void testNoFilterNoProjection() {
+    BigQueryClient bigQueryClient = mock(BigQueryClient.class);
+    prepareCalculateSizeNoFilter(bigQueryClient);
+    BigQueryEstimatedTableStatistics s =
+        BigQueryEstimatedTableStatistics.newBuilder(TABLE, bigQueryClient).build();
+    assertThat(s.numRows().isPresent()).isTrue();
```
emkornfield, line 54: might be nice to adapt Truth's [Optional Matcher](https://github.com/google/truth/blob/master/extensions/java8/src/main/java/com/google/common/truth/OptionalSubject.java) to a new matcher that works with OptionalLong

connector/src/test/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatisticsTest.java
```diff
@@ -0,0 +1,124 @@
...
+    assertThat(s.numRows().getAsLong()).isEqualTo(10L);
+    assertThat(s.sizeInBytes().isPresent()).isTrue();
+    assertThat(s.sizeInBytes().getAsLong()).isEqualTo(1000L);
```
emkornfield, line 57: it would be nice to add some comments to these assertions documenting the expected math.

emkornfield: same goes for all of these tests.

emkornfield (changes requested):
I really think doing a full query should be optional.  If we can't get the data from a dry-run query we should consider by default returning unknown.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
davidrabinowitz: I need TableStatistics in order to return two values in case a filter is involved, as I want to have just one query job

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
davidrabinowitz: Good point

davidrabinowitz:
Notice that actual queries are used in two cases:
* When using filters on tables, in order to get the number of rows. This option is not being guarded now.
* When reading views, in order to get the number of rows. This is an opt-in feature, as any read of views requires settings the `viewsEnabled` option to true.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
anoopj: +1. I think this query should have two input stages (one with pushdown and one without). Would be interesting to compare plans/perf.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
anoopj: In addition to the opt-in/opt out, it may also make sense to look at the basic table stats and filters to figure out ways to avoid running a query, whenever possible. For instance, if the table is tiny (say well under 10 MB, the Spark default broadcast join threshold), we could just report the full table size. OTOH, if the table is very large and if there are filters on clustering/partition keys, we can estimate the filter ratio.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -168,32 +171,57 @@ String fullTableName(TableId tableId) {
...
+  }
+
+  private FieldValueList queryForSingleRow(String sql) {
+    TableResult result = query(sql);
```
anoopj, line 224: Can we add a tag so that users can get visibility on queries BQ connector runs for stats?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
```diff
@@ -0,0 +1,107 @@
...
+import com.google.cloud.bigquery.connector.common.TableStatistics;
+import com.google.common.collect.ImmutableList;
+import org.apache.spark.sql.sources.v2.reader.Statistics;
+
+import java.util.Optional;
+import java.util.OptionalLong;
+
+class BigQueryEstimatedTableStatistics implements Statistics {
+
+  private static BigQueryEstimatedTableStatistics UNKNOWN_STATISTICS =
```
anoopj, line 32: This seems unused with the new logic to compute stats for views.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
davidrabinowitz: Replaced with `format("SELECT COUNT(*), COUNTIF(%s) from `%s`", f, table)`

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
davidrabinowitz: view usage already requires setting `viewsEnabled` to true. See [here](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/9c8afb6a102fcf027ef371609de5cd39872919dc/connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java#L80)

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
emkornfield: > OTOH, if the table is very large and if there are filters on clustering/partition keys, we can estimate the filter ratio.

I think this could be handled by a dry-run?

emkornfield:
> When using filters on tables, in order to get the number of rows. This option is not being guarded now.
If this isn't a dry run, I think it should be guarded (and return unknown) when not available.  Per Anoop's comments above there might be some heuristics that we could use in the unknown case without running the query (i.e. table size is less then X, just use statistics as is)

> When reading views, in order to get the number of rows. This is an opt-in feature, as any read of views requires settings the viewsEnabled option to true

For this case can we materialize the view first and then get the statistics directly, instead of running two queries?  If not then i think a second guard would be useful, since in theory HMS could be used to cache statistics for the view.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
davidrabinowitz: @emkornfield 

> I think this could be handled by a dry-run?

Unfortunately no

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
davidrabinowitz: Yes. I've added `origin=spark-bigquery-connector` to the query job

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
davidrabinowitz: It will be once we will add the guards...

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryEstimatedTableStatistics.java
davidrabinowitz: How about `Factory` and `create()` ?

emkornfield: SGTM

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
abmodi: Are we targeting these statistics for BHJ conversion? Can we just leverage spark's estimation logic here for filters and aggregates instead of running the query?

emkornfield: I believe we are taking a different approach now and trying to use data from the Read API directly.   We might want to close this PR?

davidrabinowitz: I agree. Let's reopen when relevant

PR #283, opened 2020-12-11
Implementing Datasource v2 writing support

davidrabinowitz:


connector/src/integration-test/java/com/google/cloud/spark/bigquery/v2/SparkBigQueryConnectorIntegrationTest.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.spark.bigquery.v2;
+
+
```
anoopj, line 18: Nit - please remove the blank lines. Should we add a checkstyle rule to flag the build? Also, IntelliJ autoformatting may help.

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
```diff
@@ -0,0 +1,287 @@
...
+          fields.stream().filter(field -> field.getType() != Schema.Type.NULL).findFirst();
+      return actualType.orElseThrow(
+          () -> new IllegalArgumentException("No actual type has been found in " + avroType));
+    } else {
+      return avroType;
+    }
+  }
+
+  static Converter createConverterFor(DataType sparkType, Schema avroType) {
+    if (sparkType instanceof NullType && avroType.getType() == Schema.Type.NULL) {
```
anoopj, line 144: Please feel free to ignore this, but there is a bunch of type checks and casts in the code. This is where Scala pattern matching shines, but since this is Java, I was wondering if using a visitor pattern would make this cleaner.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/AvroIntermediateRecordWriter.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.avro.io.EncoderFactory;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+public class AvroIntermediateRecordWriter implements IntermediateRecordWriter {
+
+  private final OutputStream outputStream;
+  private final DatumWriter<GenericRecord> writer;
+  // private final BinaryEncoder encoder;
```
anoopj, line 33: Remove the commented out code?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/IntermediateRecordWriter.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package com.google.cloud.spark.bigquery.v2;
+
+import org.apache.avro.generic.GenericRecord;
+import org.apache.spark.sql.catalyst.InternalRow;
+
+import java.io.IOException;
+
+public interface IntermediateRecordWriter {
```
anoopj, line 23: This should probably extend `Closeable` so that clients can use try.. with resources.

connector/src/test/java/com/google/cloud/spark/bigquery/BinaryIteratorsTest.java
```diff
@@ -0,0 +1,137 @@
+/*
```
anoopj, line 1: Probably remove this file?

connector/src/test/scala/com/google/cloud/spark/bigquery/it/AdHocSuite.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  def readAllTypesTable(dataSourceFormat: String): DataFrame =
+    spark.read.format(dataSourceFormat)
+      .option("dataset", testDataset)
+      .option("table", ALL_TYPES_TABLE_NAME)
+      .load()
+
+  test("test ad hoc") {
+    spark.conf.set("temporaryGcsBucket", "davidrab-sandbox")
+//    val s = spark
```
anoopj, line 98: Remove?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/AvroIntermediateRecordWriter.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public void close() throws IOException {
+    try {
+      // encoder.flush();
+      dataFileWriter.flush();
+    } finally {
+      dataFileWriter.close();
+      // outputStream.close();
```
anoopj, line 57: Shouldn't we close the `outputStream`?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceReader.java
```diff
@@ -85,11 +85,13 @@ public BigQueryDataSourceReader(
         new ReadSessionCreator(readSessionCreatorConfig, bigQueryClient, bigQueryReadClientFactory);
     this.globalFilter = globalFilter;
     this.schema = schema;
-    this.fields =
+    // We want to keep the key order
+    this.fields = new LinkedHashMap<>();
+    for (StructField field :
         JavaConversions.asJavaCollection(
```
anoopj, line 91: Java `Collection` doesn't guarantee ordering. Should we use a something like a list? (Using something like `JavaConversions .seqAsJavaList`)

anoopj (commented):
Initial set of feedback. I'm only half way through.

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
```diff
@@ -0,0 +1,287 @@
...
+import org.apache.avro.Conversions;
+import org.apache.avro.LogicalTypes;
+import org.apache.avro.Schema;
+import org.apache.avro.SchemaBuilder;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.util.Utf8;
+import org.apache.spark.sql.catalyst.InternalRow;
+import org.apache.spark.sql.catalyst.expressions.SpecializedGetters;
+import org.apache.spark.sql.catalyst.util.ArrayData;
+import org.apache.spark.sql.types.*;
```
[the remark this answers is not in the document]
davidrabinowitz: Yep, unfortunately intellij is to blame... I'll reconfigure it

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
```diff
@@ -0,0 +1,287 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Optional;
+
+public class AvroSchemaConverter {
```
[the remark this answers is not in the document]
davidrabinowitz: The main issue is we want a clean Java implementation in order to avoid scala versions issue (connector 2.11 on Spark 2.12 and vice versa). I'll add a test comparing the two.

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
```diff
@@ -0,0 +1,287 @@
...
+      return fieldsAssembler.endRecord();
+    }
+    if (dataType instanceof UserDefinedType) {
+      DataType userDefinedType = ((UserDefinedType) dataType).sqlType();
+      return sparkTypeToRawAvroType(userDefinedType, recordName, builder);
+    }
+    if (dataType instanceof MapType) {
+      throw new IllegalArgumentException(SchemaConverters.MAPTYPE_ERROR_MESSAGE);
+    } else {
+      throw new IllegalArgumentException("Data type not expected: " + dataType.simpleString());
```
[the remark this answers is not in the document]
davidrabinowitz: fixed

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
davidrabinowitz: I agree the code is cumbersome, but as it is built on Scala classes, meant to be used with pattern matching, they do not support the visitor pattern. Also specifically here, there's a dual matching - both to spark (catalyst) type and the avro type

connector/src/main/java/com/google/cloud/spark/bigquery/v2/AvroIntermediateRecordWriter.java
davidrabinowitz: fixed

connector/src/main/java/com/google/cloud/spark/bigquery/v2/AvroIntermediateRecordWriter.java
davidrabinowitz: fixed

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceReader.java
davidrabinowitz: fixed

connector/src/main/java/com/google/cloud/spark/bigquery/v2/IntermediateRecordWriter.java
davidrabinowitz: Good point

connector/src/test/scala/com/google/cloud/spark/bigquery/it/AdHocSuite.scala
davidrabinowitz: The entire file was added by mistake

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
```diff
@@ -0,0 +1,287 @@
...
+      return createConverterFor(userDefinedType.sqlType(), avroType);
+    }
+    throw new IllegalArgumentException(
+        String.format("Cannot convert Catalyst type %s to Avro type %s", sparkType, avroType));
+  }
+
+  @FunctionalInterface
+  interface Converter {
+    // void convert(InternalRow fromRow, GenericData.Record toRecord, int index);
+    Object convert(SpecializedGetters getters, int ordinal);
```
anoopj, line 246: Should we have a generic interface for type safety?

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
```diff
@@ -0,0 +1,287 @@
...
+      UserDefinedType userDefinedType = (UserDefinedType) sparkType;
+      return createConverterFor(userDefinedType.sqlType(), avroType);
+    }
+    throw new IllegalArgumentException(
+        String.format("Cannot convert Catalyst type %s to Avro type %s", sparkType, avroType));
+  }
+
+  @FunctionalInterface
+  interface Converter {
+    // void convert(InternalRow fromRow, GenericData.Record toRecord, int index);
```
anoopj, line 245: Remove?

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
```diff
@@ -0,0 +1,287 @@
...
+    if (sparkType instanceof ArrayType && avroType.getType() == Schema.Type.ARRAY) {
+      DataType et = ((ArrayType) sparkType).elementType();
+      boolean containsNull = ((ArrayType) sparkType).containsNull();
+
+      Converter elementConverter =
+          createConverterFor(et, resolveNullableType(avroType.getElementType(), containsNull));
+      return (getter, ordinal) -> {
+        ArrayData arrayData = getter.getArray(ordinal);
+        int len = arrayData.numElements();
+        Object[] result = new Object[len];
```
anoopj, line 215: We are allocating the Object array in the inner loop. This may cause a lot of allocations and will likely result in GC. Is there a way to avoid this by doing this lazily?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceV2.java
```diff
@@ -54,4 +65,50 @@ private SparkSession getDefaultSparkSessionOrCreate() {
...
     return createReader(null, options);
   }
+
+  @Override
+  public Optional<DataSourceWriter> createWriter(
+      String writeUUID, StructType schema, SaveMode mode, DataSourceOptions options) {
+    Injector injector =
+        createInjector(
+            schema, options, new BigQueryDataSourceWriterModule(writeUUID, schema, mode));
+    // first verify if we need to do anything at all, based on the table existence and teh save
```
anoopj, line 75: Minor: Typo in "the save"

connector/src/main/java/com/google/cloud/spark/bigquery/v2/IntermediateDataCleaner.java
```diff
@@ -0,0 +1,64 @@
...
+package com.google.cloud.spark.bigquery.v2;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Responsible for recursively deleting the intermediate path. Implementing Runnable in order to act
```
anoopj, line 25: Not sure if this is a good idea to run data cleanup as a shutdown hook. Generally shutdown hooks should run quickly and it's entirely possible that the GCS clients and associated thread pools may be in the process of shutting down, so it's hard to guarantee that this cleanup actually happened.

Should we consider alternatives? For instance, can we upload the temporary files with random prefixes and use GCS lifecycle policies to automatically clean them up after a safe window?

anoopj (commented):
Some more feedback.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/ParquetIntermediateRecordWriter.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.avro.generic.GenericRecord;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.avro.AvroParquetWriter;
+import org.apache.parquet.hadoop.ParquetWriter;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+class ParquetIntermediateRecordWriter implements IntermediateRecordWriter {
```
anoopj, line 28: Do we have some benchmarks on how Parquet vs Avro for intermediate data? My guess is that writing Parquet is going to be much more expensive and was wondering if it made sense, given that it's a short-lived intermediate format.

davidrabinowitz: Based on the DSv1 data, parquet was faster for data sizes up to 250 GB or so, from there Avro was faster. I'll run a comparison test

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
davidrabinowitz: We are allocating the result array - this array is then wrapped by Arrays.asList(), serving as the backing data structure of the list. This code is a direct 1:1 port of the spark-avro scala code

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
```diff
@@ -0,0 +1,287 @@
...
+        fieldsAssembler.name(field.name()).type(avroType).noDefault();
+      }
+      return fieldsAssembler.endRecord();
+    }
+    if (dataType instanceof UserDefinedType) {
+      DataType userDefinedType = ((UserDefinedType) dataType).sqlType();
+      return sparkTypeToRawAvroType(userDefinedType, recordName, builder);
+    }
+    if (dataType instanceof MapType) {
+      throw new IllegalArgumentException(SchemaConverters.MAPTYPE_ERROR_MESSAGE);
```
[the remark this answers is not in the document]
davidrabinowitz: It means that we will need to iterate the schema twice, once for validation and once for conversion. I agree this is BigQuery specific, but at this point this is used only at the BQ connector and tailored to its needs. A more generic implementation exist in the spark-avro project.

connector/src/main/java/com/google/cloud/spark/bigquery/AvroSchemaConverter.java
davidrabinowitz: No - the only place where it is used is in `GenericData.Record.put()` which accepts `Object`. There is no meaning for generics here.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceReaderModule.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Optional;
+
+public class BigQueryDataSourceReaderModule implements Module {
+  @Override
+  public void configure(Binder binder) {
+    // empty
+  }
+
+  @Singleton
```
functicons, line 36: Why is this a singleton? What if the process reads 2 tables?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceV2.java
```diff
@@ -15,35 +15,46 @@
...
 import org.apache.spark.sql.sources.v2.ReadSupport;
+import org.apache.spark.sql.sources.v2.WriteSupport;
 import org.apache.spark.sql.sources.v2.reader.DataSourceReader;
+import org.apache.spark.sql.sources.v2.writer.DataSourceWriter;
 import org.apache.spark.sql.types.StructType;
 
 import java.util.Optional;
 
-public class BigQueryDataSourceV2 implements DataSourceV2, ReadSupport {
+public class BigQueryDataSourceV2 implements DataSourceV2, ReadSupport, WriteSupport {
```
functicons, line 39: Add some high level comments about the implementation.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceV2.java
```diff
@@ -54,4 +65,50 @@ private SparkSession getDefaultSparkSessionOrCreate() {
   public DataSourceReader createReader(DataSourceOptions options) {
     return createReader(null, options);
   }
+
+  @Override
+  public Optional<DataSourceWriter> createWriter(
```
functicons, line 70: Add comments about when it will be empty.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceWriterModule.java
```diff
@@ -0,0 +1,120 @@
...
+    this.sparkSchema = sparkSchema;
+    this.mode = mode;
+  }
+
+  @Override
+  public void configure(Binder binder) {
+    // empty
+  }
+
+  @Singleton
```
functicons, line 54: The same question.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceWriterModule.java
```diff
@@ -0,0 +1,120 @@
...
+                                      ".spark-bigquery-%s-%s", applicationId, UUID.randomUUID()));
+                      return String.format(
+                          "gs://%s/%s", config.getPersistentGcsBucket().get(), path);
+                    });
+        gcsPath = new Path(gcsPathOption);
+        FileSystem fs = gcsPath.getFileSystem(conf);
+        needNewPath = fs.exists(gcsPath); // if the path exists for some reason, then retry
+      }
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
```
functicons, line 116: Why use UncheckedIOException?

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryIndirectDataSourceWriter.java
```diff
@@ -0,0 +1,276 @@
...
+import java.io.UncheckedIOException;
+import java.util.AbstractMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class BigQueryIndirectDataSourceWriter implements DataSourceWriter {
```
functicons, line 58: Add comments about what "Indirect" means.

functicons (changes requested):
As a general comment, 1) add comments for all public classes and methods; 2) break down a huge PR into multiple PRs, which will be much easier for review.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/IntermediateDataCleaner.java
davidrabinowitz: This is done just as a safe-guard in case `BigQueryIndirectDataSourceWriter.cleanTemporaryGcsPathIfNeeded` has not been called by an error. Relying on GCS object lifecycle will require the bucket to allow lifecycle management, which is not guaranteed.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/ParquetIntermediateRecordWriter.java
davidrabinowitz: After running a benchmark, it appears that the Parquet writer as constantly slower than the avro one, so I've removed it.

connector/src/integration-test/java/com/google/cloud/spark/bigquery/v2/SparkBigQueryConnectorIntegrationTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.spark.sql.types.ShortType;
+import org.apache.spark.sql.types.StringType;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+import org.apache.spark.sql.types.TimestampType;
+import org.apache.spark.sql.types.UserDefinedType;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+
+public class SparkBigQueryConnectorIntegrationTest {
```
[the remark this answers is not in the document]
davidrabinowitz: It was added by mistake.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceReaderModule.java
davidrabinowitz: Spark create a new `DataSourceReader`/`Writer` for each read.

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceWriterModule.java
davidrabinowitz: `provideDataSourceWriter()` now throws

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceWriterModule.java
davidrabinowitz: The `BigQueryDataSourceWriter` is created once per write operation, so it is singleton within this context

functicons (approved):

PR #286, opened 2020-12-17
Improved filter and pseudo column support

esert-g:
This PR introduces some changes to connector's filter and pseudo column handling so that it can take advantage of the BQ Storage API. 

Filter changes: 
- Removed push down filter restrictions based on readDataFormat (i.e. both Avro and Arrow behave the same). 
- Added support for EqualNullSafe filter type (which can't be supported natively by the BQ Storage API, but it can be supported as a compound expression). 
- Removed restriction on struct field filtering (which is supported by BQ Storage API), but spark itself doesn't seem to try push them anyways at the moment.

Pseudo column changes: If queried BQ table is [ingestion time partitioned](https://cloud.google.com/bigquery/docs/creating-partitioned-tables), schema of the table is modified with the additions of top level `_PARTITIONTIME` and `_PARTITIONDATE` fields. This introduces a slight behavior change for `select *` queries (i.e. the two new pseudo columns are returned on top of regular schema). This can't be avoided at the moment because there isn't a way to differentiate queries within connector as far as I can tell.

On top of the unit tests, I also ran some tests on a 2.4 dataproc cluster (specifically on pseudo columns).

### DIFF
--- a/connector/src/main/java/com/google/cloud/bigquery/connector/common/LoggingBigQueryStorageReadRowsTracer.java
+++ b/connector/src/main/java/com/google/cloud/bigquery/connector/common/LoggingBigQueryStorageReadRowsTracer.java
@@ -94,7 +94,7 @@ public class LoggingBigQueryStorageReadRowsTracer implements BigQueryStorageRead
       return 0;
     }
     Duration time = timer.getAccumulatedTime();
-    double seconds =  (time.toMillis() / 1000.0);
+    double seconds = (time.toMillis() / 1000.0);
     if (seconds != 0) {
       return (long) (metric / seconds);
     }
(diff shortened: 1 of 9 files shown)


davidrabinowitz (changes requested):
One of the integration tests had failed:
```
Step #2 - "integration-tests": [info] - user defined schema. DataSource com.google.cloud.spark.bigquery.v2.BigQueryDataSourceV2 *** FAILED ***
Step #2 - "integration-tests": [info]   StructType(StructField(whatever,ByteType,true)) did not equal StructType(StructField(word,StringType,false), StructField(word_count,LongType,false), StructField(corpus,StringType,false), StructField(corpus_date,LongType,false)) (SparkBigQueryEndToEndITSuite.scala:398)
Step #2 - "integration-tests": [info]   Analysis:
Step #2 - "integration-tests": [info]   StructType(0: StructField(whatever,ByteType,true) -> StructField(word,StringType,false), 1: -> StructField(word_count,LongType,false), 2: -> StructField(corpus,StringType,false), 3: -> StructField(corpus_date,LongType,false))
```
Can you please check it out?

esert-g:
Closing this in favor of https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/375

PR #292, opened 2021-01-03
New Notebook Examples

cpcgoogle:
Adding three examples of using Spark and BigQuery.


davidrabinowitz (changes requested):
Can you please remove the `.DS_Store` from the commit?

Issue #248, opened 2020-09-30
Writes for Dataframes with more than 10000 partitions fail


davidrabinowitz:
When trying to write dataframes with more than 10000 partitions the load job fails with a  "Too many sources provided: XXXX. Limit is 10000" error. This should be fixed with a better URI provided to the load job.

anton-statutov-booking:
Hello, I just stumbled upon on this problem. Do you know a good workaround?

davidrabinowitz:
Fixed in version 0.18.1

anton-statutov-booking:
Thanks. I've tested it - the fix works.

asiunov:
This solution in https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/295 may not work in all cases, e.g. when there are many skipped partitions (because *usually* spark doesn't write empty parts). The easiest workaround is to do `df.coalesce(10000).write...`.

yeshvantbhavnasi:
Even I see the error popping on with latest version of bigquery-connector

PR #295, opened 2021-01-11
Issue #248: reducing the size of the URI list for bq load

davidrabinowitz:


connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
```diff
@@ -122,4 +128,45 @@ public static String friendlyTableName(TableId tableId) {
...
+   * @param prefixRegex the regex to catch the part before the file index.
+   * @param suffixRegex the regex to catch the part after the file index.
+   */
+  public static List<String> optimizeLoadUriList(
+      List<String> uris, String prefixRegex, String suffixRegex) {
+    Pattern pattern = Pattern.compile("(" + prefixRegex + "\\d*)\\d\\**(" + suffixRegex + ")");
+    ImmutableList.Builder<String> result = ImmutableList.builder();
+    List<String> workingList = uris;
+    Multimap<String, String> trimmedUriMap = HashMultimap.create();
+    while (!workingList.isEmpty()) {
```
functicons, line 147: Can we introduce a helper method trimUris which would be cleaner?

while (true) {
  trimmedUris = trimUris(currentUris);
  if (trimmedUris.size() == currentUris.size()) {
     // Couldn't reduce further.
    break;
  }
  currentUri = trimmedUril;
}

connector/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryUtil.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.google.cloud.spark.bigquery;
+
+import com.google.cloud.bigquery.connector.common.BigQueryUtil;
+
+import java.util.List;
+
+/** Spark related utilities */
+public class SparkBigQueryUtil {
+  public static List<String> optimizeLoadUriListForSpark(List<String> uris) {
```
functicons, line 24: Add a comment about this method.

functicons (changes requested):

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
davidrabinowitz: Notice that in case the collection of the mapped URIs contains less than 10 URIs then a wildcard is not being used and they are added to the result list. Moving the entire logic into the helper method would mean to process those several times in case of a large result. I took your advice for cleaning the method and moved the trimming part to another method.

connector/src/test/java/com/google/cloud/spark/bigquery/OptimizeLoadUriListTest.java
```diff
@@ -0,0 +1,107 @@
...
+  private List<String> input;
+  private List<String> expected;
+
+  public OptimizeLoadUriListTest(List<String> input, List<String> expected) {
+    this.input = input;
+    this.expected = expected;
+  }
+
+  @Parameterized.Parameters(name = "{index}: Should get {1}")
+  public static Iterable<Object[]> data() {
```
functicons, line 41: Can we add edge cases for input list of size 0 and 1?

functicons (approved):

connector/src/test/java/com/google/cloud/spark/bigquery/OptimizeLoadUriListTest.java
davidrabinowitz: Added, thanks

Issue #296, opened 2021-01-14
Dependency conflict with slf4j


npetunin-clgx:
When we are using this library, in Spark job, we see the following conflict: 
```
Caused by: java.lang.LinkageError: loader constraint violation in interface itable initialization: when resolving method "com.google.cloud.spark.bigquery.BigQueryUtilScala$.org$apache$spark$internal$Logging$$log_()Lorg/slf4j/Logger;" the class loader (instance of org/apache/spark/util/ChildFirstURLClassLoader) of the current class, com/google/cloud/spark/bigquery/BigQueryUtilScala$, and the class loader (instance of sun/misc/Launcher$AppClassLoader) for interface org/apache/spark/internal/Logging have different Class objects for the type org/slf4j/Logger used in the signature 
```
We found that we have extra slf4j dependency in the bigquery connector dependency tree in Bigquery Connector version: 0.18.0. We tried to exclude slf4j dependency from `arrow-vector`, build it and try customized jar. And we didn't see error specified above. 

We would like to exclude `slf4j-api` dependency from `arrow-vector` as follows:
```
"org.apache.arrow" % "arrow-vector" % "0.16.0" exclude("org.slf4j", "slf4j-api"), 
```
in Line 82 in `build.sbt` file.

stsai-clgx:
hi @davidrabinowitz, 
thank you for resolving this issue.  When will next release become available?

PR #298, opened 2021-01-15
Issue #296: Removed redundant packaged slf4j-api

davidrabinowitz:


build.sbt
```diff
@@ -79,7 +79,7 @@ lazy val connector = (project in file("connector"))
       "aopalliance" % "aopalliance" % "1.0" % "provided",
       "org.codehaus.jackson" % "jackson-core-asl" % "1.9.13" % "provided",
       "org.codehaus.jackson" % "jackson-mapper-asl" % "1.9.13" % "provided",
-      "org.apache.arrow" % "arrow-vector" % "0.16.0",
+      "org.apache.arrow" % "arrow-vector" % "0.16.0" exclude("org.slf4j", "slf4j-api"),
```
bsidhom, line 82: Is it more appropriate to mark `slf4j` as `provided` instead? Is this included by another transitive dependency and somehow not resolved consistently by SBT, or is some external distribution already including it?

davidrabinowitz: We have it as provided, unfortunately SBT still brings it from Arrow (the resolution algorithm is unlike Maven's) . This is why we need the explicit exclude.

bsidhom (approved):
LGTM

build.sbt
bsidhom: Yeah, now I can see that _just_ above the fold where the diff was actually shown. :)

PR #305, opened 2021-02-06
Adding read from query capability

davidrabinowitz:
Adding the option to run an ad-hoc query on BigQuery and then read the result directly into a dataframe. The syntax is in the following manner:
```
df = df = spark.read.format("bigquery")
      .option("viewsEnabled", true)
      .option("materializationDataset", <dataset>)
      .load("SELECT ...")
```
or
```
df = df = spark.read.format("bigquery")
      .option("viewsEnabled", true)
      .option("materializationDataset", <dataset>)
      .option("query", "SELECT ...")
      .load()
```
The queries are materialized to a temporary table and are re-used if the same query is used. For example
`df.select("col1").where("condition")` should use the temporary table and not re-run the query.

This also gives the option to efficiently run complex queries using joins on BigQuery specific functions from spark.

### DIFF
--- a/connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
+++ b/connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
@@ -40,7 +40,6 @@ import org.slf4j.LoggerFactory;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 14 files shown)


connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -216,4 +252,99 @@ public long calculateTableSize(TableInfo tableInfo, Optional<String> filter) {
           "Querying table size was interrupted on the client side", e);
     }
   }
+
+  public TableInfo materializeQueryToTable(String querySql, int viewExpirationTimeInHours) {
```
functicons, line 256: Add comments to all public methods.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -216,4 +252,99 @@ public long calculateTableSize(TableInfo tableInfo, Optional<String> filter) {
...
+    TableId tableId = createDestinationTable(Optional.empty(), Optional.empty());
+    return materializeTable(querySql, tableId, viewExpirationTimeInHours);
+  }
+
+  public TableInfo materializeViewToTable(
+      String querySql, TableId viewId, int viewExpirationTimeInHours) {
+    TableId tableId =
+        createDestinationTable(
```
functicons, line 264: I don't quite understand why we need to convert viewId to tableId here, a comment would be helpful.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -216,4 +252,99 @@ public long calculateTableSize(TableInfo tableInfo, Optional<String> filter) {
...
+  private TableInfo materializeTable(
+      String querySql, TableId destinationTableId, int viewExpirationTimeInHours) {
+    try {
+      return destinationTableCache.get(
+          querySql,
+          new DestinationTableBuilder(
+              this, querySql, destinationTableId, viewExpirationTimeInHours));
+    } catch (ExecutionException e) {
+      throw new BigQueryConnectorException(
+          BIGQUERY_VIEW_DESTINATION_TABLE_CREATION_FAILED, "Error creating destination table", e);
```
functicons, line 278: Should we add `querySql` in the error message for debugging purpose?

connector/src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndReadFromQueryITSuite.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import com.google.cloud.bigquery._
+import com.google.cloud.spark.bigquery.TestUtils
+import org.apache.spark.sql.SparkSession
+import org.scalatest.concurrent.TimeLimits
+import org.scalatest.prop.TableDrivenPropertyChecks
+import org.scalatest.{BeforeAndAfter, BeforeAndAfterAll, FunSuite, Matchers}
+
+
+class SparkBigQueryEndToEndReadFromQueryITSuite extends FunSuite
```
functicons, line 26: What if the query fails? Can you add a test case for it?

functicons (changes requested):

functicons (approved):

PR #312, opened 2021-02-23
Bumping the log severity from debug in case of an optimized count() run

davidrabinowitz:
Based on conversation with @emkornfield


connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryDataSourceReader.java
```diff
@@ -228,6 +232,7 @@ private ReadRowsResponseToInternalRowIteratorConverter createConverter(
 
   List<InputPartition<InternalRow>> createEmptyProjectionPartitions() {
     long rowCount = bigQueryClient.calculateTableSize(tableId, globalFilter);
+    logger.info("Creating a DataFrame of empty rows with size " + rowCount);
```
emkornfield, line 235: Could we phrase this as something like "Used optimized BQ count(*) path. Count: "

bsidhom (approved):
LGTM

emkornfield (approved):
Thanks!

PR #313, opened 2021-02-23
Solving Scala 2.12 private API dependency

davidrabinowitz:
Taking `SerializableConfiguration `from Apache-Beam rather than from a spark private API. This had worked with Scala 2.11 but fails with Scala 2.12.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -75,6 +75,9 @@ lazy val connector = (project in file("connector"))
       "org.apache.spark" %% "spark-core" % sparkVersion % "provided",
       "org.apache.spark" %% "spark-sql" % sparkVersion % "provided",
       "org.apache.spark" %% "spark-mllib" % sparkVersion % "provided",
+      "org.apache.beam" % "beam-sdks-java-io-hadoop-common" % "2.28.0"
+        exclude("org.apache.beam", "beam-sdks-java-core")
+        exclude("org.checkerframework", "checker-qual"),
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -75,6 +75,9 @@ lazy val connector = (project in file("connector"))
       "org.apache.spark" %% "spark-core" % sparkVersion % "provided",
       "org.apache.spark" %% "spark-sql" % sparkVersion % "provided",
       "org.apache.spark" %% "spark-mllib" % sparkVersion % "provided",
+      "org.apache.beam" % "beam-sdks-java-io-hadoop-common" % "2.28.0"
+        exclude("org.apache.beam", "beam-sdks-java-core")
+        exclude("org.checkerframework", "checker-qual"),
```
bsidhom, line 80: I saw the comment on the PR itself, but could you explain why we're grabbing this from Beam rather than Spark? Did Beam just reproduce the same code that's used internally by Spark? Should we just copy the necessary class(es) into this package?

davidrabinowitz: Hadoop's configuration is not `Serializable`. Spark has its own, but it is marked as spark private using Scala. While the Scala 2.11 compiler does not force private classes when they  are imported in Java code, it appears that the Scala 2.12 does force that. This apache beam common provides the same functionality in a public class, and it does not add any additional dependencies to the connector.

bsidhom: Thanks for clarifying.

bsidhom (approved):
LGTM

PR #314, opened 2021-02-24
Parametrised JavaShakespeare.java example

aag-peace:
Modified JavaShakespeare.java to incorporate parameters for input/output/temporary GCS buckets

### DIFF
--- a/connector/src/main/java/com/google/cloud/spark/bigquery/examples/JavaShakespeare.java
+++ b/connector/src/main/java/com/google/cloud/spark/bigquery/examples/JavaShakespeare.java
@@ -60,10 +60,10 @@ public class JavaShakespeare {
             .option("table", inputBigqueryTable)
             .load()
             .cache();
-    wordsDF.createOrReplaceTempView("words");
 
     wordsDF.show();
     wordsDF.printSchema();
+    wordsDF.createOrReplaceTempView("words");
 
     // Perform word count.
     Dataset<Row> wordCountDF =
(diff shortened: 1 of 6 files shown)


connector/src/main/java/com/google/cloud/spark/bigquery/examples/JavaShakespeare.java
```diff
@@ -28,16 +29,38 @@ public static void main(String[] args) {
...
+            inputBigqueryTable = arg.substring(2);
+            break;
+          case 'o':
+            outputBigqueryTable = arg.substring(2);
+            break;
+          default:
+            usage();
+        }
+      }
+    }
```
medb, line 51: I think it will be simpler just to hard code positional parameters:

```java
inputBigqueryTable = arg[0];
outputBigqueryTable = arg[1];
```

Also, probably we do not need to parametrize input table, because it's publicly accessible anyway, so parametrization could be as simple as:

```java
String outputBigqueryTable = arg[0];
```

medb (changes requested):

connector/src/main/java/com/google/cloud/spark/bigquery/examples/JavaShakespeare.java
aag-peace: Done

medb (approved):

aag-peace:
Closing this, due to unrelated changes

PR #319, opened 2021-02-24
Dependencies version update

davidrabinowitz:


build.sbt
```diff
@@ -16,9 +16,9 @@
 lazy val scala211Version = "2.11.12"
 lazy val scala212Version = "2.12.10"
 lazy val sparkVersion = "2.4.0"
-lazy val grpcVersion = "1.30.2"
+lazy val grpcVersion = "1.35.0"
```
medb, line 19: 1.36.0 version already released.

medb (approved):

build.sbt
davidrabinowitz: I know, the grpc version is taken from the bigquery clients for compatibility

PR #338, opened 2021-03-11
Add basic instrumentation for ReadRows calls

emkornfield:
For now I've only instrumented the Arrow DSv2 path.  As needed we can potentially add more paths and collect different metrics.

This is a strawman proposoal, happy to make changes if we think there is a better approach.

CC @davidrabinowitz


connector/src/main/java/com/google/cloud/spark/bigquery/v2/ArrowInputPartition.java
```diff
@@ -15,9 +15,7 @@
  */
 package com.google.cloud.spark.bigquery.v2;
 
-import com.google.cloud.bigquery.connector.common.BigQueryReadClientFactory;
-import com.google.cloud.bigquery.connector.common.ReadRowsHelper;
-import com.google.cloud.bigquery.connector.common.ReadSessionResponse;
+import com.google.cloud.bigquery.connector.common.*;
```
davidrabinowitz, line 18: Please avoid `.*` imports

davidrabinowitz (approved):

connector/src/main/java/com/google/cloud/spark/bigquery/v2/ArrowInputPartition.java
emkornfield: sorry about that, this is a fresh intellij install.  Is there a style file I should import?

davidrabinowitz: Not yet unfortunately... Once we will have the new architecture and the move to maven I will integrate a [google-java-format](https://github.com/google/google-java-format#third-party-integrations) maven plugin.

Issue #340, opened 2021-03-12
Compiling connector on java 11 throws exception `java.lang.UnsupportedOperationException`.


maasg:
As reported in #200 and #290, the project fails to compile against Java 11 and requires Java 8 for compilation.

davidrabinowitz:
As written [here](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/290#issuecomment-749280836), 

> For the moment, please compile the connector with Java 8. Java 11 compilation support is on our roadmap.

maasg:
@davidrabinowitz I opened a new ticket to create a PR against it. Here: https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/341
That patch seems to work well.
Is there something more to consider to support Java 11?
We are actually blocked by the old arrow dependency in the project. Using Java 11 provides a low friction upgrade.

Issue #290, opened 2020-12-22
Compiling connector on java 11 throws exception `java.lang.UnsupportedOperationException`.


lucasdavid:
Good evening.

I'm having the same issue described in #200, even though I updated to the latest version at github/releases page.
I then proceeded to compile (in the hope it would fix it), but I'm receiving a similar error message after executing `sbt assembly`:
```shell
[info] - compare arrow and avro results *** FAILED ***
[info]   java.lang.UnsupportedOperationException: sun.misc.Unsafe or java.nio.DirectByteBuffer.<init>(long, int) not available
```

Please advise me on how to move forward.

Longer shell output chunks of interest:
```shell
USER@MACHINE:~/repos/spark-bigquery-connector$ javac --version
javac 11.0.9.1

USER@MACHINE:~/repos/spark-bigquery-connector$ sbt assembly
[info] Loading project definition from /home/USER/repos/spark-bigquery-connector/project/project
[info] Loading project definition from /home/USER/repos/spark-bigquery-connector/project
[info] Compiling 1 Scala source to /home/USER/repos/spark-bigquery-connector/project/target/scala-2.10/sbt-0.13/classes...
	[info] Set current project to root (in build file:/home/USER/repos/spark-bigquery-connector/)
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by sbt.ivyint.ErrorMessageAuthenticator$ (file:/home/USER/.sbt/boot/scala-2.10.7/org.scala-sbt/sbt/0.13.17/ivy-0.13.17.jar) to field java.net.Authenticator.theAuthenticator
WARNING: Please consider reporting this to the maintainers of sbt.ivyint.ErrorMessageAuthenticator$
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
[info] Compiling 7 Java sources to /home/USER/repos/spark-bigquery-connector/published/target/scala-2.11/test-classes...
[info] Including from cache: scala-library-2.11.12.jar
[info] Compiling 12 Scala sources and 36 Java sources to /home/USER/repos/spark-bigquery-connector/connector/target/scala-2.11/classes...
[info] ScalaTest
[info] Run completed in 34 milliseconds.
[info] Total number of tests run: 0
[info] Suites: completed 0, aborted 0
[info] Tests: succeeded 0, failed 0, canceled 0, ignored 0, pending 0
[...]

PR #341, opened 2021-03-12
Add runtime flag to allow usage of unsafe API in compile/tests

maasg:
This PR addresses the compilation issue related to netty/arrow as documented in #290 

> For Java 11, -Dio.netty.tryReflectionSetAccessible=true is required additionally for Apache Arrow library. This prevents java.lang.UnsupportedOperationException: sun.misc.Unsafe or java.nio.DirectByteBuffer.(long, int) not available when Apache Arrow uses Netty internally.

Adding the compiler flag and `fork`-ing the jvm for testing solves the problem.

That process broke the `BigQueryUtilsSuite.scala` which is made more robust in this PR.


maasg:
fixes #340

emkornfield:
LGTM for the unsafe change, don't know scala well enough to understand the changes there.

davidrabinowitz:
We are mandating Java 8 since PR #371

PR #353, opened 2021-03-29
issue# 190 : supported writing data to BQ with description

himanshukohli09:


connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryIndirectDataSourceWriter.java
```diff
@@ -264,17 +262,29 @@ void updateMetadataIfNeeded() {
...
+        newField.setDescription(marker);
+      } else if (!description.endsWith(marker)) {
+        newField.setDescription(description + " " + marker);
+      }
     }
     return newField.build();
   }
 
+  private boolean doesFieldContainsDescription(StructField field) {
+    return (field.metadata().contains("description")
```
davidrabinowitz, line 284: Please use the built in `StructField.getComment()`

davidrabinowitz (changes requested):

connector/src/main/java/com/google/cloud/spark/bigquery/v2/BigQueryIndirectDataSourceWriter.java
himanshukohli09: Here I am reading from the key "description" of the Metadata, whereas StructField.getComment() reads from the key "comment" of the Metadata. 
So we intend to support both the keys? or just one of the keys?
 if we intend to support both the keys and if user supplies different values to both the keys what shall we do?

davidrabinowitz: It was my mistake to force the BigQuery name "description" to Spark, we should support the standard `getComment()`. For backwards compatibility we will support both, where the standard  `getComment()` takes precedence.

davidrabinowitz: Notice the comment field should also be populated when reading from BigQuery

himanshukohli09: Done

davidrabinowitz (changes requested):

davidrabinowitz (approved):

connector/src/main/java/com/google/cloud/spark/bigquery/SchemaConverters.java
```diff
@@ -276,15 +276,21 @@ protected static Field createBigQueryColumn(StructField sparkField, int depth) {
...
+    if (description.isPresent()) fieldBuilder.setDescription(description.get());
+
+    return fieldBuilder.build();
+  }
 
-    return createBigQueryFieldBuilder(fieldName, fieldType, fieldMode, subFields)
-        .setDescription(description)
-        .build();
+  public static Optional<String> getDescriptionOrCommentOfField(StructField field) {
+    if (!field.getComment().isEmpty()) return Optional.of(field.getComment().get());
```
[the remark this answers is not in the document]
himanshukohli09: Done

PR #355, opened 2021-03-31
Upgrade arrow 3

emkornfield:
Note, on my machine I see some failures for deduplication in SBT for one specific version of scala when running "sbt assembly", I'm not sure if this is local to my machine, I got exclusions wrong or something else needs to be changed with SBT.  I would appreciate it some hints if this doesn't pass CI.


davidrabinowitz (approved):

maasg:
@emkornfield This is great! Thanks!

emkornfield:
> @emkornfield This is great! Thanks!

Was the arrow version causing you any issues?  It shouldn't be exposed publicly (it should be shaded)

PR #357, opened 2021-04-06
Add configuration options.

emkornfield:
Adds options for:
- Endpoint of the Read API - Useful for testing and if we migrate to regional endpoints
- Encoded base CreateReadSessionRequest which allows for setting hidden fields if necessary
- Arrow compression codec - This isn't hooked up yet in Arrow code (we need to wait until Arrow 4.0)

Additionally
- Fix some formatting from prior commits.
- Force version of guava to not be android (otherwise compilation/intellij are flaky)
- Introduce new configuration object for ReadRows calls

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,7 +52,6 @@ lazy val commonTestDependencies = Seq(
   "io.grpc" % "grpc-alts" % grpcVersion exclude("io.grpc", "grpc-netty-shaded"),
   "io.grpc" % "grpc-netty" % grpcVersion,
   "com.google.api" % "gax-grpc" % "1.60.0" exclude("io.grpc", "grpc-netty-shaded"),
-  "com.google.guava" % "guava" % "30.0-jre",
 
   "org.scalatest" %% "scalatest" % "3.1.0" % "test",
   "org.mockito" %% "mockito-scala-scalatest" % "1.10.0" % "test",
@@ -91,6 +90,7 @@ lazy val connector = (project in file("connector"))
       "aopalliance" % "aopalliance" % "1.0" % "provided",
       "org.codehaus.jackson" % "jackson-core-asl" % "1.9.13" % "provided",
       "org.codehaus.jackson" % "jackson-mapper-asl" % "1.9.13" % "provided",
+      "com.google.guava" % "guava" % "30.1.1-jre",
       "com.google.inject" % "guice" % "4.2.3",
       "org.apache.arrow" % "arrow-vector" % "4.0.0"
 			  excludeAll(ExclusionRule(organization="org.slf4j"),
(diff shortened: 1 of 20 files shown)


emkornfield:
@davidrabinowitz could you provide logs of gcb run failure?

emkornfield:
@davidrabinowitz I removed the optional changes (I moved the comment to right above where the optional fields start and add serializability unit tests).

emkornfield:
@davidrabinowitz could you provide logs from build failures?

emkornfield:
@davidrabinowitz could you share latest failures from gcbrun?

emkornfield:
@davidrabinowitz builds are now successful.

build.sbt
```diff
@@ -16,7 +16,7 @@
 lazy val scala211Version = "2.11.12"
 lazy val scala212Version = "2.12.10"
 lazy val sparkVersion = "2.4.0"
-lazy val grpcVersion = "1.30.2"
+lazy val grpcVersion = "1.35.0"
```
davidrabinowitz, line 19: please revert the grpc/netty/bigquery versions to those of version 0.19.1 ? When those were updated in 0.19.0 it caused problems to several users. We will update those after the v1 release

davidrabinowitz (changes requested):

build.sbt
emkornfield: Do you know exactly which library this was for?  I was hoping it was netty which I haven't touched here.  Do you have a repro of the issue to see if these changes are safe?

The bigquery/gax/grpc  updates are driven by needing the latest version of the proto files to support compression configuration.  Users can always use the serialized proto approach but I'd prefer to leave it undocumented.

Any thoughts on the other changes?

emkornfield: (I tried just keeping the storage library and tests fail with this approach).

davidrabinowitz: grpc-netty has a dependency on netty

emkornfield: reverted.

emkornfield:
Ping @davidrabinowitz

emkornfield:
ping @davidrabinowitz

build.sbt
```diff
@@ -91,6 +90,7 @@ lazy val connector = (project in file("connector"))
       "aopalliance" % "aopalliance" % "1.0" % "provided",
       "org.codehaus.jackson" % "jackson-core-asl" % "1.9.13" % "provided",
       "org.codehaus.jackson" % "jackson-mapper-asl" % "1.9.13" % "provided",
+      "com.google.guava" % "guava" % "30.1.1-jre",
```
davidrabinowitz, line 93: Please return guava to commonTestDependencies, as it is used by the acceptance test in `published`

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryReadClientFactory.java
```diff
@@ -15,22 +15,29 @@
  */
 package com.google.cloud.bigquery.connector.common;
 
+import com.esotericsoftware.minlog.Log;
```
davidrabinowitz, line 18: I don't see it used in the code. Why do we need an additional logger?

davidrabinowitz (changes requested):

build.sbt
emkornfield: done.

connector/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryReadClientFactory.java
emkornfield: removed.  intellij be only open.

davidrabinowitz (approved):

PR #372, opened 2021-04-23
Add QueueingInputPartitionReader for parallel processing of Readers.

emkornfield:
I'll send out a PR incorporating this into the reader separately.

Ran thread dependendent/timing tests 1000 times to help verify
no issues.


emkornfield:
Realized this is at the wrong level.  Needs to be for ArrowReader.  Going to close and reopen when i fix that.

Issue #383, opened 2021-04-30
Failed to update table metadata when writing to a partitioned table


bskim45:
Hi, I recently updated version to `0.20.0`, but getting following errors when I try to write to a partitioned table.

example code:
```
df
    .write
    .format("bigquery")
    .option("datePartition", "20210430")
    .option("partitionType", "DAY")
    .option("createDisposition", "CREATE_IF_NEEDED")
    .option("intermediateFormat", "orc")
    .mode("overwrite")
    .save("awesome_project.dataset_name.table_name")
```

error:
```
Exception in thread "main" java.lang.RuntimeException: Failed to write to BigQuery
	at com.google.cloud.spark.bigquery.BigQueryWriteHelper.writeDataFrameToBigQuery(BigQueryWriteHelper.scala:94)
	at com.google.cloud.spark.bigquery.BigQueryInsertableRelation.insert(BigQueryInsertableRelation.scala:43)
	at com.google.cloud.spark.bigquery.BigQueryRelationProvider.createRelation(BigQueryRelationProvider.scala:113)
	at org.apache.spark.sql.execution.datasources.SaveIntoDataSourceCommand.run(SaveIntoDataSourceCommand.scala:46)
	at org.apache.spark.sql.execution.command.ExecutedCommandExec.sideEffectResult$lzycompute(commands.scala:70)
	at org.apache.spark.sql.execution.command.ExecutedCommandExec.sideEffectResult(commands.scala:68)
	at org.apache.spark.sql.execution.command.ExecutedCommandExec.doExecute(commands.scala:90)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$execute$1(SparkPlan.scala:180)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:218)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
	at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:215)
	at org.apache.spark.sql.execution.SparkPlan.execute(SparkPlan.scala:176)
	at org.apache.spark.sql.execution.QueryExecution.toRdd$lzycompute(QueryExecution.scala:124)
	at org.apache.spark.sql.execution.QueryExecution.toRdd(QueryExecution.scala:123)
[...]

PR #385, opened 2021-05-02
Fixed table metadata update when writing to a partitioned table (Issue #383)

davidrabinowitz:


connector/src/test/scala/com/google/cloud/spark/bigquery/it/SparkBigQueryEndToEndWriteITSuite.scala
```diff
@@ -318,6 +318,47 @@ class SparkBigQueryEndToEndWriteITSuite extends FunSuite
...
+           |('2020-07-02', 'bar')
+           |""".stripMargin.replace('\n', ' ')))
+
+      // overrding a single partition
+      val newDataDF = spark.createDataFrame(
+        List(Row(java.sql.Date.valueOf("2020-07-01"), "baz")).asJava,
+        StructType(Array(
+          StructField("the_date", DateType).withComment("the partition field"),
+          StructField("some_text", StringType))))
+
```
himanshukohli09, line 348: This test and the previous test ("overwrite single partition" : starting from line 279) are almost same, with the only difference being the "val newDataDF". Can we reuse the code by moving the common reusable code to some method?

himanshukohli09 (commented):
Suggesting a change, please have a look.

himanshukohli09 (approved):
LGTM

PR #402, opened 2021-05-28
Fix next() to be interrupt safe.

emkornfield:
for some reason I did not think Spark would interrupted the thread calling
the next() method, I had trouble tracking this down but thinking more thorougly
about it I think spark could and probably does do this.  In this case
if the thread was interrupted while trying to get the next batch it is
possible for a memory leak to occur.

This also rearranges the close order for DSv2, so that everything shutdown more cleanly, this also could avoid whatever edge case we are seeing with memory leaks.

Unfortunately, the unit test change didn't replicate the flakiness, but
I still think these changes make the code safer.


emkornfield:
DO NOT MERGE WITH https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/402/commits/a71607da0629203d8212d70844858498d4e79e84 in place

emkornfield:
@davidrabinowitz gcbrun passed for me three times in a row now.

davidrabinowitz (approved):

PR #416, opened 2021-06-17
Moved zstd-jni library to be provided in order to solve Spark 2.4 compatibility

davidrabinowitz:
When using Arrow compression in Spark 2.4 the user need to add `--conf spark.executor.userClassPathFirst=true --conf spark.driver.userClassPathFirst=true --packages com.github.luben:zstd-jni:1.4.9-1` to the spark-submit

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,9 +95,12 @@ lazy val connector = (project in file("connector"))
         exclude("org.checkerframework", "checker-qual"),
       "org.slf4j" % "slf4j-api" % "1.7.16" % "provided",
       "aopalliance" % "aopalliance" % "1.0" % "provided",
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -221,14 +226,17 @@ val excludedOrgs = Seq(
 lazy val renamed = Seq(
   "avro.shaded",
   "com.fasterxml",
+  // "com.github.luben",
```
himanshukohli09, line 229: since this is commented, can we delete this line?

himanshukohli09 (changes requested):

himanshukohli09 (approved):

PR #419, opened 2021-06-21
Fix bug where background threads > 2 cases would miss pages

emkornfield:
- Simplify threaded model for multiplexer and parallel arrow reader.
  This uses the same model that gRPC users where all messages for
  consumers go through the same queue and a sentiel/exception is used
  placed in the queue to notify of an error. This is a simplication and
  is not involved in the bug fix. A semaphore is introduced to prevent
  the queues from filling.

- For parallel arrow reader.  Create the batch as part of the task
  submitted to the thread pool, this eliminates complicated future
  handling code.

- In ArrowColumnBatchPartitionReader don't instantiate SequenceInputStream
  unless it is needed.  SequenceInputStream preconsumes records so for
  the multithreaded case pages are lost (actual bug fix).

### DIFF
--- a/connector/src/main/java/com/google/cloud/bigquery/connector/common/IteratorMultiplexer.java
+++ b/connector/src/main/java/com/google/cloud/bigquery/connector/common/IteratorMultiplexer.java
@@ -120,7 +120,7 @@ public class IteratorMultiplexer<T> implements AutoCloseable {
 
     @Override
     public T next() {
-      Preconditions.checkState(t != TERMINAL_SENTINEL, "next element cannot be terminal");
+      Preconditions.checkState(t != TERMINAL_SENTINEL, "No next message");
       if (t instanceof RuntimeException) {
         throw (RuntimeException) t;
       }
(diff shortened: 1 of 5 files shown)


connector/src/main/java/com/google/cloud/bigquery/connector/common/IteratorMultiplexer.java
```diff
@@ -96,46 +96,50 @@ void readAhead() {
...
-      return !done.get() || !queue.isEmpty();
+      return t != TERMINAL_SENTINEL;
     }
 
     @Override
     public T next() {
-      Preconditions.checkState(t != null, "next element cannot be null");
-      if (rethrow != null) {
-        throw rethrow;
+      Preconditions.checkState(t != TERMINAL_SENTINEL, "next element cannot be null");
```
davidrabinowitz, line 123: Error message should be "next element cannot be the sentinel"

davidrabinowitz (approved):

connector/src/main/java/com/google/cloud/bigquery/connector/common/IteratorMultiplexer.java
emkornfield: changed to "no next message"

emkornfield:
@davidrabinowitz let me know if the failure is related to my changes, and I can split them.

PR #425, opened 2021-06-25
Make background processing classes more robust to errors

emkornfield:
If a throwable error is made (for instance ZSTD error) previously this would result in a hang because it wasn't caught and handled in the submitted task.  This PR make changes to catch throwables everyplace necessary to avoid issues like this in the future.


davidrabinowitz (changes requested):
Which Errors does ZSTD throws? Can we make sure that we are not swallowing `Error`s needlessly (such as OOM)?

emkornfield:
> Which Errors does ZSTD throws? Can we make sure that we are not swallowing Errors needlessly (such as OOM)?

Note this change doesn't swallow the errors.  It makes sure they are all caught and surfaced to user facing threads.  The issue before is a background thread would die with one of these exceptions, and user facing threads would block endlessly waiting for data.

davidrabinowitz (approved):

PR #438, opened 2021-08-09
Spark DataSource V2 Write with "BigQuery Storage Write API"

himanshukohli09:


himanshukohli09:
> Question - will the Storage Write API support the allowFieldAddition and allowFieldRelaxation options?
> 
> This is supported as part of the BigQuery load API, but not as part of Storage Write API. So I think it would need to be done explicitly by the connector (check schema, if needed update it prior to write).

We are currently evaluating this.

PR #439, opened 2021-08-11
Re-architected acceptance tests

davidrabinowitz:


davidrabinowitz:
Aborted in favor of #440

PR #458, opened 2021-09-14
Update google-cloud-bigquerystorage dependency to 2.1.2 in branch-0.22

esert-g:
This required me to update grpcVersion as well.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,7 +16,7 @@
 lazy val scala211Version = "2.11.12"
 lazy val scala212Version = "2.12.10"
 lazy val sparkVersion = "2.4.0"
-lazy val grpcVersion = "1.37.1"
+lazy val grpcVersion = "1.40.1"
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -135,7 +135,7 @@ lazy val connector = (project in file("connector"))
 
       // Keep com.google.cloud dependencies in sync
       "com.google.cloud" % "google-cloud-bigquery" % "1.131.1",
```
davidrabinowitz, line 137: Shouldn't google-cloud-bigquery be upgraded to 2.1.7 or is it an unrelated change?

esert-g: google-cloud-bigquerystorage update to 2.1.2 is only necessary for handling some new RESOURCE_EXHAUSTED errors that may be sent from the bq storage read api calls. I don't know why we would need to update google-cloud-bigquery for that reason.

PR #460, opened 2021-09-17
Allow Reading of External Tables in Spark Connector

anoopj:
Testing: existing unit testing pass, end-to-end tests


davidrabinowitz (changes requested):
Can you please add an integration test testing a read from an external table?

davidrabinowitz (changes requested):
Can you please add an integration test testing a read from an external table?

davidrabinowitz (approved):

PR #480, opened 2021-10-19
Update pom.xml upgrade gcs-connector version to 2.2.3

ismailsimsek:


davidrabinowitz:
@ismailsimsek thanks for the update, however notice it applies only when running integration tests outside of the GCP environment. It does not apply for production installation as the connector uses the GCS connector installed in the cluster - it does not bring one of it's own.

PR #509, opened 2021-12-16
Log4j 1.2 removal from SchemaConverterTest.java

davidrabinowitz:


VinayLondhe14 (approved):

shadiramadan:
@davidrabinowitz I'm assuming this is for the log4j exploit.  Do you know when this will be released?

davidrabinowitz:
The library does not use log4j in its code. There were two tests that had imported log4j directly, and those were removed from the code.

davidrabinowitz:
Soon, however the released connector has no runtime dependency on log4j (1.2 or 2.x) - it was used only in the tests

PR #546, opened 2022-02-10
Simplified logic of creating gcsPath and fixed infinite loop bug by checking if persistentGcsPath already exists

### LINKED ISSUE #544: df.write hangs indefinitely if persistentGcsPath already exists
Hello,

I just had a massive headache trying to figure out why my Spark job was hanging indefinitely and failing to run any tasks. I discovered that if you are specifying a `persistentGcsBucket` and `persistentGcsPath` and the `persistentGcsPath` already exists on the GCS bucket, the call to `df.write` will hang indefinitely without raising an error. This occurs even if the destination directory is completely empty.

This is using `gs://spark-lib/bigquery/spark-bigquery-with-dependencies_2.12-0.23.2.jar`

Example code to reproduce:

```
data = [
    (1, 'A'),
    (2, 'B'),
]
df = spark.createDataFrame(data=data, schema=['id', 'val'])
df.write.format("bigquery").option(
    "persistentGcsBucket", "some-bucket"
).option(
    "persistentGcsPath", "test"
).mode("overwrite").save("somedataset.sometable")
```
Now, you should see output in `gcs://some-bucket/test/`. Delete the contents of that directory, then rerun:
```
df.write.format("bigquery").option(
    "persistentGcsBucket", "some-bucket"
).option(
    "persistentGcsPath", "test"
).mode("overwrite").save("somedataset.sometable")
```

This will hang indefinitely.

VinayLondhe14:
Already removed while loop

Fixes #544


davidrabinowitz (changes requested):
The code assumes that having a UUID ensures that the directory does not exist. While statistically it's correct, we can't count on it, and that's why we need the `while(needNewPath) if a UUID is involved. 

Also, please have a single util method for this in java (in spark-bigquery-connector-common), and call it from both v1 and v2 code

VinayLondhe14:
Thank you for the review. 

I will create a single util method for this in Java and call it from both v1 and v2.

Given that the chances of UUIDs colliding are astronomically low ([Link 1](https://stackoverflow.com/questions/24876188/how-big-is-the-chance-to-get-a-java-uuid-randomuuid-collision), [Link 2](https://stackoverflow.com/questions/2513573/how-good-is-javas-uuid-randomuuid)), I'd argue that the while loop makes the code less readable and error-prone for a negligible benefit and would favor removing it altogether. Let me know if that is an acceptable tradeoff.

VinayLondhe14:
@davidrabinowitz Created a single util method and added a while loop as suggested

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryUtil.java
```diff
@@ -48,4 +53,61 @@ private static Properties loadBuildProperties() {
...
+   * @param applicationId A unique identifier for the Spark application
+   * @return org.apache.hadoop.fs.Path object backed by GCS
+   * @throws IOException
+   */
+  public static Path createGcsPath(
+      SparkBigQueryConfig config, Configuration conf, String applicationId) throws IOException {
+    Path gcsPath;
+    Preconditions.checkArgument(
+        config.getTemporaryGcsBucket().isPresent() || config.getPersistentGcsBucket().isPresent(),
+        "Temporary or persistent GCS bucket must be informed.");
```
davidrabinowitz, line 73: Please change the message to "Either Temporary or persistent GCS bucket must be set"

davidrabinowitz (changes requested):

spark-bigquery-dsv2/spark-bigquery-dsv2-common/src/main/java/com/google/cloud/spark/bigquery/v2/context/BigQueryDataSourceWriterModule.java
```diff
@@ -103,35 +104,39 @@ public BigQueryIndirectDataSourceWriterContext provideIndirectDataSourceWriterCo
 
   Path createGcsPath(SparkBigQueryConfig config, Configuration conf, String applicationId)
```
davidrabinowitz, line 105: Method should removed in favor of hte `SparkBigQueryUtil` one

davidrabinowitz (changes requested):

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryUtil.java
VinayLondhe14: Done

spark-bigquery-dsv2/spark-bigquery-dsv2-common/src/main/java/com/google/cloud/spark/bigquery/v2/context/BigQueryDataSourceWriterModule.java
VinayLondhe14: Ah good catch

PR #553, opened 2022-02-25
Exclude com.github.luben.zstd from being shaded

VinayLondhe14:
According to https://github.com/luben/zstd-jni#limitations, the zstd-jni library should not be relocated. 

With this change, ZSTD Arrow compression can now be used without running into NoClassDefFoundError

Thank you @emkornfield!

### DIFF
--- a/spark-bigquery-dsv1/spark-bigquery-with-dependencies-parent/pom.xml
+++ b/spark-bigquery-dsv1/spark-bigquery-with-dependencies-parent/pom.xml
@@ -50,6 +50,7 @@
             <relocation>
               <pattern>com.github</pattern>
               <shadedPattern>com.google.cloud.spark.bigquery.repackaged.com.github</shadedPattern>
+              <!-- Need to exclude the zstd library from being repackaged because of https://github.com/luben/zstd-jni#limitations -->
               <excludes>
                 <exclude>com.github.luben.zstd.**</exclude>
               </excludes>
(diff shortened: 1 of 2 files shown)


spark-bigquery-dsv1/spark-bigquery-with-dependencies-parent/pom.xml
```diff
@@ -50,6 +50,9 @@
             <relocation>
               <pattern>com.github</pattern>
               <shadedPattern>com.google.cloud.spark.bigquery.repackaged.com.github</shadedPattern>
+              <excludes>
```
emkornfield, line 53: is it possible to check if we can just remove this pattern altogether?  If not please add a comment on why we are doing the exclusion.

emkornfield (changes requested):

spark-bigquery-dsv1/spark-bigquery-with-dependencies-parent/pom.xml
VinayLondhe14: As of now we can remove the pattern altogether. My only concern with that is if start using another library with the `com.github` pattern in the future and accidentally include it here, the ZSTD implementation will break again. Excluding it specifically is much safer in that case. 

Let me know if that makes sense. Otherwise, happy to remove the pattern completely as well.

emkornfield: That is fair, I think the probability of this is very low.  We could ensure things don't break by having appropriate integration tests.

emkornfield (approved):
This is fine for now.  I think there should be a follow-up to have appropriate end-to-end test to ensure now regressions.

VinayLondhe14:
Yes. Agreed. I will create a follow-up bug to write appropriate tests for this.

PR #561, opened 2022-03-14
Added traceId tracking

davidrabinowitz:


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
```diff
@@ -365,6 +369,16 @@ public static SparkBigQueryConfig from(
               + config.cacheExpirationTimeInMinutes);
     }
 
+    String traceIdParam =
+        getAnyOption(globalOptions, options, "traceId").or(SparkBigQueryUtil.getJobId(sqlConf));
+    if (traceIdParam.length() > MAX_TRACE_ID_LENGTH) {
```
emkornfield, line 374: I think we do bytes, so this would this should probably encode in UTF-8 before checking if you want to be strict.

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
```diff
@@ -365,6 +369,16 @@ public static SparkBigQueryConfig from(
               + config.cacheExpirationTimeInMinutes);
     }
 
+    String traceIdParam =
+        getAnyOption(globalOptions, options, "traceId").or(SparkBigQueryUtil.getJobId(sqlConf));
```
emkornfield, line 373: should defaulting to getting the Job ID be something that is opt-in?

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
```diff
@@ -365,6 +369,16 @@ public static SparkBigQueryConfig from(
...
+    if (traceIdParam.length() > MAX_TRACE_ID_LENGTH) {
+      throw new IllegalArgumentException(
+          String.format(
+              "traceId cannot longer than %d. Provided value was [%s]",
+              MAX_TRACE_ID_LENGTH, traceIdParam));
+    } else {
```
emkornfield, line 379: not sure if there is a standard in code, but the else isn't necessary here.

emkornfield (commented):
A few questions and comments worth discussing.  Is code-cov setup correctly?  it looks like tests for ReadSessionCreator might not be sufficient?

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
davidrabinowitz: The [client API](https://github.com/googleapis/java-bigquerystorage/blob/v2.11.0/proto-google-cloud-bigquerystorage-v1/src/main/java/com/google/cloud/bigquery/storage/v1/ReadSession.java#L5878) requests String

emkornfield: right, but the String will be converted to UTF-8 first, so the string length here will not be accurate for non-ascii characters.  I think maybe either remove this check or do the check?

VinayLondhe14 (approved):

PR #563, opened 2022-03-16
making sure that in savemode==append the table is not deleted on abort

davidrabinowitz:


spark-bigquery-dsv2/spark-bigquery-dsv2-common/src/main/java/com/google/cloud/spark/bigquery/v2/context/BigQueryDirectDataSourceWriterContext.java
```diff
@@ -50,7 +50,7 @@
   private final String writeUUID;
   private final RetrySettings bigqueryDataWriterHelperRetrySettings;
 
-  private final TableId temporaryTableId;
+  private final BigQueryTable writtenTable;
```
VinayLondhe14, line 53: Minor: Can we change the name to something like `tableToWrite`? Since it is possible that the table does not exist and `writtenTable` implies that we have already written something to it

spark-bigquery-dsv2/spark-bigquery-dsv2-common/src/main/java/com/google/cloud/spark/bigquery/v2/context/BigQueryDirectDataSourceWriterContext.java
```diff
@@ -108,7 +108,7 @@ public BigQueryDirectDataSourceWriterContext(
    * @return The TableId to which Spark will do the writing: whether that is the destinationTableID
    *     or the temporaryTableId.
    */
-  private TableId getOrCreateTable(
```
VinayLondhe14, line 111: Also, while we are on this file, can we change the above comment to 
```
This function determines whether the destination table exists: if it doesn't, we will create a table and Spark will directly write to it
...
```

VinayLondhe14 (approved):

VinayLondhe14 (approved):

PR #570, opened 2022-03-22
Refactored dsv1 read path to use ReadSessionCreator

VinayLondhe14:


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkFilterUtils.java
```diff
@@ -321,6 +321,10 @@ static String compileValue(Object value, char arrayStart, char arrayEnd) {
     return value.toString();
   }
 
+  public static Optional<String> emptyIfNeeded(String value) {
```
davidrabinowitz, line 324: should move to BigQueryUtil (no spark dependency)

davidrabinowitz (changes requested):

davidrabinowitz (approved):

PR #588, opened 2022-04-09
Implement helper classes and utility functions for query pushdown

VinayLondhe14:

PR #589, opened 2022-04-09
Implement converter from Spark expressions to SQL and inject it into BigQueryStrategy

VinayLondhe14:
Builds on #588


davidrabinowitz (changes requested):
Please add license headers where it is missed

davidrabinowitz (approved):

PR #592, opened 2022-04-13
CVE-2021-35515, CVE-2021-35516, CVE-2021-35517, CVE-2021-36090: upgrading commons-compress

davidrabinowitz:


spark-bigquery-parent/pom.xml
```diff
@@ -244,6 +244,11 @@
                 <artifactId>arrow-vector</artifactId>
                 <version>${arrow.version}</version>
             </dependency>
+            <dependency>
```
VinayLondhe14, line 247: Is this a transitive dependency that we are fixing to a particular version?

davidrabinowitz: transitive from arrow, even in arrow 7 they haven't upgraded commons-compress, and we are getting security alerts on it

VinayLondhe14: Gotcha

VinayLondhe14 (approved):

PR #598, opened 2022-04-21
Handle the exception in load when job is not created

irajhedayati:
I added a check to see if the `finishedJob` is `null`. If it is, then just re-throw the exception thrown by the `google-cloud-bigquery` library.

### DIFF
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
@@ -514,6 +514,10 @@ public class BigQueryClient {
             finishedJob.getJobId());
       }
     } catch (Exception e) {
+      if (finishedJob == null) {
+        log.error("Unable to create the job to load to {}", BigQueryUtil.friendlyTableName(options.getTableId()));
+        throw e;
+      }
       TimePartitioning.Type partitionType = options.getPartitionTypeOrDefault();
 
       if (e.getMessage()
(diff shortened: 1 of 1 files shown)


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -514,6 +514,10 @@ public void loadDataIntoTable(
             finishedJob.getJobId());
       }
     } catch (Exception e) {
+      if (finishedJob == null) {
```
VinayLondhe14, line 517: Can we move this below the if loop just before `JobId jobId = finishedJob.getJobId();`?

irajhedayati: I initially put it there. Then, checking the logic, the catch clause is implemented when the job is created successfully but failed to finish. Hence, if the `finishedJob` is `null`, then there is no need to check the exception message against the partition type. That's why I moved it to the type so other developers understand its priority.

VinayLondhe14: ah. understood. good point

VinayLondhe14 (approved):

VinayLondhe14:
Can you run `mvn spotless:apply` and update the PR? It has some format violations currently

irajhedayati:
Done that @VinayLondhe14 ... first commit :)

VinayLondhe14:
Thank you!

PR #602, opened 2022-04-22
changes to support cacheExpirationTimeInMinutes=0, if user wants to disable cache

suryasoma:
Fix to the issue: https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/593

changes to support cacheExpirationTimeInMinutes=0, if user wants to disable cache
Added unit test
Ran integration tests

### DIFF
--- a/spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/SparkBigQueryConfigTest.java
+++ b/spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/SparkBigQueryConfigTest.java
@@ -191,6 +191,29 @@ public class SparkBigQueryConfigTest {
   }
 
   @Test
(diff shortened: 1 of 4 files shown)


spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/SparkBigQueryConfigTest.java
```diff
@@ -190,6 +190,29 @@ public void testConfigFromOptions() {
     assertThat(config.getBigQueryJobLabels()).containsEntry("foo", "bar");
   }
 
+  @Test
```
VinayLondhe14, line 193: Can you add another test to make sure an exception is thrown if a negative value is passed? Optionally, you can also add a test to make sure that a positive value passes

VinayLondhe14 (approved):

PR #611, opened 2022-05-03
Update CHANGES.md and README.md to capture the next release's info

askdre:
During a release, automation will create two commits.

The release commit will
* replace the ## Next header in CHANGES.md with the release version and date
* remove the ## Unreleased Changes section from README.md
* replace ${next-release-tag} in README.md with the release version

The next development commit will
* Add the ## Next section to CHANGES.md
* Replace the release README.md with the development README.md containing the ## Unreleased Changes section and ${next-release-tag} template.
* Update the link to the latest release in the README.md to point to the current release version.

### DIFF
--- a/README-template.md
+++ b/README-template.md
@@ -1,0 +1,1140 @@
+# Apache Spark SQL connector for Google BigQuery
+
+<!--- TODO(#2): split out into more documents. -->
+
+The connector supports reading [Google BigQuery](https://cloud.google.com/bigquery/) tables into Spark's DataFrames, and writing DataFrames back into BigQuery.
+This is done by using the [Spark SQL Data Source API](https://spark.apache.org/docs/latest/sql-programming-guide.html#data-sources) to communicate with BigQuery.
+
+## Unreleased Changes
+
+This Readme may include documentation for changes that haven't been released yet.  The latest release's documentation and source code are found here.
+
+https://github.com/GoogleCloudDataproc/spark-bigquery-connector/tree/0.24.2
+
+## BigQuery Storage API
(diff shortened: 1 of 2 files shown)


davidrabinowitz (changes requested):
The README is the main documentation document for the users, and as such it needs to contain the last released version in plain-text, not tag. If memory serves, the original idea was to have a README development doc with the tags in another directory. The release process will copy and transform this doc to the final README in the project root

askdre:
Moved the templated readme to README-template.md which will get instantiated to generate the README.md.  Developers should be notified that any changes made to README.md will be overriden by changes to README-template.md during the release, so they should add their documentation updates to README-template.md.

README-template.md
```diff
@@ -0,0 +1,1140 @@
...
+  .save("dataset.table")
+```
+
+Partition overwriting and the use of `datePartition`, `partitionField` and `partitionType` as described below is not
+supported at this moment by the direct write method.
+
+**Important:** Please refer to the [data ingestion pricing](https://cloud.google.com/bigquery/pricing#data_ingestion_pricing)
+page regarding the BigQuery Storage Write API pricing.
+
+**Important:** Please use version ${next-release-tag} and above for direct writes, as previous
```
davidrabinowitz, line 252: Here the version should be the constant 0.24.2, due to the PR #563 fix

davidrabinowitz (changes requested):

README-template.md
askdre: Updated line 252 to use the constant instead of the varible.

davidrabinowitz (approved):

PR #620, opened 2022-05-12
Created a Spark expression factory to avoid binary compatibility issues

VinayLondhe14:
Ran into an issue where one of the Spark expressions called `Alias` is not binary compatible between Spark versions. See https://stackoverflow.com/questions/66106096/delta-lake-insert-into-sql-in-pyspark-is-failing-with-java-lang-nosuchmethoder

So, created a Spark expression factory that can create Alias for specific Spark versions

Note that even though there are 18 changed files, most of the changes are pretty small. Basically, changing the method parameters


suryasoma (approved):
LGTM

PR #621, opened 2022-05-13
Integrated SparkExpressionFactory with BigQueryStrategy

VinayLondhe14:

PR #622, opened 2022-05-13
Integrated SparkPlanFactory with BigQueryStrategy

VinayLondhe14:
This PR is built on top of #621 and injects the SparkPlanFactory class into BigQueryStrategy.

This is done to extract out the logic of creating the SparkPlan in a separate class and to facilitate easier unit testing


suryasoma (approved):
LGTM

PR #625, opened 2022-05-17
Add minor fixes encountered during testing

VinayLondhe14:
Encountered a few things while testing that need fixing

### DIFF
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
@@ -22,7 +22,6 @@ import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.Clustering;
 import com.google.cloud.bigquery.Dataset;
 import com.google.cloud.bigquery.DatasetId;
-import com.google.cloud.bigquery.DatasetInfo;
 import com.google.cloud.bigquery.FormatOptions;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.JobConfiguration;
@@ -140,18 +139,6 @@ public class BigQueryClient {
   public TableInfo createTable(TableId tableId, Schema schema) {
     TableInfo tableInfo = TableInfo.newBuilder(tableId, StandardTableDefinition.of(schema)).build();
     return bigQuery.create(tableInfo);
-  }
-
-  /**
-   * Creates a Dataset in BigQuery
-   *
-   * @return the created Dataset
-   */
-  public Dataset createDataset() {
-    String dataset =
-        String.format("spark_bigquery_%d_%d", System.currentTimeMillis(), System.nanoTime());
-    DatasetId datasetId = DatasetId.of(dataset);
-    return bigQuery.create(DatasetInfo.of(datasetId));
   }
 
   /**
@@ -308,9 +295,7 @@ public class BigQueryClient {
   TableId createDestinationTable(
       Optional<String> referenceProject, Optional<String> referenceDataset) {
     String project = materializationProject.orElse(referenceProject.orElse(null));
-    String dataset =
-        materializationDataset.orElse(
-            referenceDataset.orElse(createDataset().getDatasetId().getDataset()));
+    String dataset = materializationDataset.orElse(referenceDataset.orElse(null));
     String name =
         String.format(
             "_bqc_%s", UUID.randomUUID().toString().toLowerCase(Locale.ENGLISH).replace("-", ""));
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
@@ -144,6 +144,7 @@ public class BigQueryClient {
 
   /**
    * Creates a Dataset in BigQuery
+   *
    * @return the created Dataset
    */
   public Dataset createDataset() {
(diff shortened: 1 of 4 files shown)


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/direct/BigQueryRDDFactory.java
```diff
@@ -95,7 +95,7 @@ public RDD<InternalRow> buildScanFromSQL(String sql) {
 
     log.info(
         "Querying table {}, requiredColumns=[{}]",
-        actualTable.getFriendlyName(),
+        actualTable.getTableId().getTable(),
```
VinayLondhe14, line 98: `actualTable.getFriendlyName()` return null. `actualTable.getTableId().getTable()` returns the correct name

spark-bigquery-pushdown/pushdown_common_src/main/scala/com/google/cloud/spark/bigquery/pushdowns/BigQueryStrategy.scala
```diff
@@ -51,9 +51,6 @@ class BigQueryStrategy(expressionConverter: SparkExpressionConverter, expression
     try {
       generateSparkPlanFromLogicalPlan(plan)
     } catch {
-      case ue: BigQueryPushdownUnsupportedException =>
```
VinayLondhe14, line 54: We do not want to throw an exception here because even though pushdown is unsupported for such operations (such as `CreateViewCommand`), they will be completed in Spark. An exception terminates the entire process and we don't want that to happen.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -295,7 +307,9 @@ Iterable<Table> listTables(DatasetId datasetId, TableDefinition.Type... types) {
   TableId createDestinationTable(
       Optional<String> referenceProject, Optional<String> referenceDataset) {
     String project = materializationProject.orElse(referenceProject.orElse(null));
-    String dataset = materializationDataset.orElse(referenceDataset.orElse(null));
+    String dataset =
```
VinayLondhe14, line 310: Open question here. If customers do not pass in a dataset while pushing down a query, do we create one for them?

Here, I have added code that creates a dataset if none is given

spark-bigquery-pushdown/pushdown_common_src/main/scala/com/google/cloud/spark/bigquery/pushdowns/BigQueryStrategy.scala
davidrabinowitz: This should be in a comment. Also, if why don't we want to fall spark it we cannot generate a spark plan due to an exception ? Create view should not be pushed down

davidrabinowitz (changes requested):
I'm not sure the connector needs to create datasets

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
VinayLondhe14: Synced with David. We do not want to create the dataset ourselves for now

spark-bigquery-pushdown/pushdown_common_src/main/scala/com/google/cloud/spark/bigquery/pushdowns/BigQueryStrategy.scala
VinayLondhe14: Added a comment in the code regarding this.

Since this is an all or nothing pushdown, all the commands are tried using the `BigQueryStrategy` first. If BigQueryStrategy returns Nil, it falls back to other Spark strategies. 

We can probably add some optimizations here to fast fail if the passed in plan contains unsupported nodes

VinayLondhe14: I have added a method `hasUnsupportedNodes()` that checks for unsupported nodes in the plan. We will need to compare performance with and without the check because with the check we are essentially going through the plan twice.

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/direct/BigQueryRDDFactory.java
davidrabinowitz, line 98: This will print just the table name, not the project and dataset

spark-bigquery-pushdown/pushdown_common_src/main/scala/com/google/cloud/spark/bigquery/pushdowns/BigQueryStrategy.scala
```diff
@@ -48,18 +49,35 @@ class BigQueryStrategy(expressionConverter: SparkExpressionConverter, expression
    *         query generation was successful, None if not.
    */
   override def apply(plan: LogicalPlan): Seq[SparkPlan] = {
+    // Check if we have any unsupported nodes in the plan. If we do, we return
+    // Nil and let Spark try other strategies
+    if(hasUnsupportedNodes(plan)) {
```
davidrabinowitz, line 54: Is `order by` is going to be solved by another PR?

VinayLondhe14: yes. this wont solve the `order by` problem. the solution for that will have to be investigated

davidrabinowitz (approved):

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/direct/BigQueryRDDFactory.java
VinayLondhe14: fixed

PR #626, opened 2022-05-18
Added support to enable/disable query pushdown in Pyspark

VinayLondhe14:
Screen Shot 2022-05-17 at 5 22 37 PM


davidrabinowitz (approved):

gary-arcana:
@VinayLondhe14 , I'm curious why this isn't always enabled by default?

gary-arcana:
I see, thanks for the explanation @davidrabinowitz

PR #641, opened 2022-05-25
removing conscrypt from shaded dependencies 

suryasoma:
removing conscrypt from shaded dependencies, so that it will be part of org/conscrypt since acceptance tests were failing in serverless with reason `failed to find class org/conscrypt/CryptoUpcalls`. Anyway we are not using the package `com.google.cloud.spark.bigquery.repackaged.org.conscrypt` anywhere in the code so it is safe to remove the shaded dependency

### DIFF
--- a/spark-bigquery-dsv1/spark-bigquery-with-dependencies-parent/pom.xml
+++ b/spark-bigquery-dsv1/spark-bigquery-with-dependencies-parent/pom.xml
@@ -115,11 +115,6 @@
               </shadedPattern>
             </relocation>
             <relocation>
-              <pattern>org.conscrypt</pattern>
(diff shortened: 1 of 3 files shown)


davidrabinowitz (commented):
conscrypt is still included in the shaded jars, only in an unshaded form. Please the following entry to the `<dependencyManagement>` section of spark-bigquery-parent/pom.xml:
```
<dependency>
    <groupId>org.conscrypt</groupId>
    <artifactId>conscrypt-openjdk-uber</artifactId>
    <version>2.5.1</version>
    <scope>provided</scope>
</dependency>
```
And test again. Please also verify that conscrypt is not included in the final jar

davidrabinowitz (approved):

PR #674, opened 2022-06-16
Increasing default maxParallelism to 10K for BigLake external tables

davidrabinowitz:


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/ReadSessionCreator.java
```diff
@@ -103,8 +105,8 @@ public ReadSessionResponse create(
             .getMaxParallelism()
             .orElseGet(
                 () -> {
-                  log.debug("using default parallelism [{}]", config.getDefaultParallelism());
-                  return config.getDefaultParallelism();
+                  log.debug("using default parallelism [{}]", DEFAULT_MAX_PARALLELISM);
+                  return DEFAULT_MAX_PARALLELISM;
```
emkornfield, line 109: was this previously where we had the division logic of table size / XXX MB?

emkornfield: If it is I think we should keep that logic for the time being if we can retrieve a size and default to 10/100K if we can't  once we have a better API we can eliminate the division logic.

davidrabinowitz: Yes, it is not new

emkornfield: so are we still doing the division some place else?

suryasoma (approved):

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/ReadSessionCreator.java
```diff
@@ -124,6 +134,15 @@ public ReadSessionResponse create(
     return new ReadSessionResponse(readSession, actualTable);
   }
 
+  private int calculateDefaultMaxParallelismForTable(TableDefinition tableDefinition) {
+    if (tableDefinition instanceof StandardTableDefinition) {
+      StandardTableDefinition standardTableDefinition = (StandardTableDefinition) tableDefinition;
+      return Math.max(
+          (int) (standardTableDefinition.getNumBytes() / DEFAULT_BYTES_PER_PARTITION), 1);
```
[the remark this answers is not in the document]
davidrabinowitz: @sonatype-lift ignore

PR #687, opened 2022-07-02
Support Spark dynamic partition pruning.

coufon:
Proof-of-Concept of Spark Dynamic Partition Pruning.


coufon:
Close the PR since it is not for submission.

PR #689, opened 2022-07-07
Remove @Ignore annotation on DataSourceV1QueryPushdownIntegrationTest

VinayLondhe14:
Updated the test `testStringFunctionExpressions` to ensure it passes. 

Will need to keep an eye on `testDateFunctionExpressions` because as far as I remember, that test was failing previously but is passing now. If it fails again, I will update it

Also, changed the file name from `DatasourceV1QueryPushdownIntegrationTest` to `DataSourceV1QueryPushdownIntegrationTest` to be consistently named with other classes


suryasoma (approved):
LGTM

PR #720, opened 2022-07-27
Adding support for equal null safe operator in query pushdown

suryasoma:


spark-bigquery-tests/src/main/java/com/google/cloud/spark/bigquery/integration/QueryPushdownIntegrationTestBase.java
```diff
@@ -128,7 +128,8 @@ public void testBasicExpressions() {
                     + "word_count & corpus_date, "
                     + "word_count | corpus_date, "
                     + "word_count ^ corpus_date, "
-                    + "~ word_count "
+                    + "~ word_count, "
+                    + "word <=> corpus "
```
VinayLondhe14, line 132: It might be worth checking if this actually gets pushed down to BQ or it gets computed in Spark prior to push down.

I ask because Spark is smart enough to realize that since we have word = 'augurs' in the WHERE clause, it will never be equal to `corpus`

VinayLondhe14 (approved):

spark-bigquery-tests/src/main/java/com/google/cloud/spark/bigquery/integration/QueryPushdownIntegrationTestBase.java
suryasoma: I did check it Vinay, the plan was getting created with word <=> corpus

PR #736, opened 2022-08-11
adding a flag to assert pushdown is passed during integration tests

suryasoma:


spark-bigquery-tests/src/main/java/com/google/cloud/spark/bigquery/integration/QueryPushdownIntegrationTestBase.java
```diff
@@ -16,12 +17,15 @@ public class QueryPushdownIntegrationTestBase extends SparkBigQueryIntegrationTe
 
   @Test
   public void testStringFunctionExpressions() {
+    BigQueryConnectorUtils.enablePushdownSession(spark);
```
VinayLondhe14, line 20: Instead of enabling and disabling pushdown manually for every test, why don't we do it in a @Before and @After method?

suryasoma: addressed

davidrabinowitz (approved):

davidrabinowitz:
obselete

PR #749, opened 2022-08-25
fixing the behavior of log with pushdown to that of spark

suryasoma:


spark-bigquery-pushdown/pushdown_common_src/main/scala/com/google/cloud/spark/bigquery/pushdowns/SparkExpressionConverter.scala
```diff
@@ -253,7 +253,7 @@ abstract class SparkExpressionConverter {
         ConstantString("RAND") + ConstantString("()")
```
davidrabinowitz, line 254: I'd change the names from left and right to base and expr, In BigQuery it is [LOG(expr, base)](https://cloud.google.com/bigquery/docs/reference/standard-sql/mathematical_functions#log) whereas in spark it is [LOG(base,expr)](https://spark.apache.org/docs/latest/api/sql/#log)

davidrabinowitz (changes requested):

davidrabinowitz (approved):

PR #767, opened 2022-09-07
Adding the option createReadSessionTimeoutInSeconds, to override the timeout for CreateReadSession

suryasoma:


CHANGES.md
```diff
@@ -3,6 +3,7 @@
 ## Next
 
 * PR #745: Supporting load from query in spark-3.1-bigquery.
+* PR #767: Adding the option readSessionTimeoutInSeconds, to set the max timeout for a read session.
```
kmjung, line 6: Nit: I'd like to make it clear that this value specifies an override value for the deadline on `CreateReadSession` rather than an override for the eventual expiration time of the session. The latter is not user-configurable.

suryasoma: Okay

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientFactory.java
```diff
@@ -134,6 +138,23 @@ private BigQueryReadClient createBigQueryReadClient(Optional<String> endpoint) {
           BigQueryReadSettings.newBuilder()
               .setTransportChannelProvider(transportBuilder.build())
               .setCredentialsProvider(FixedCredentialsProvider.create(credentials));
+      if (bqConfig.getCreateReadSessionTimeoutInSeconds().isPresent()) {
+        // Setting the read session timeout only of the user provided one using options or using the
```
VinayLondhe14, line 142: Nit: only if rather than only of

VinayLondhe14 (approved):

davidrabinowitz (changes requested):

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientFactory.java
suryasoma: Addressed

davidrabinowitz (approved):

PR #770, opened 2022-09-09
Escape single quote in Contains, EndsWith and StartsWith

VinayLondhe14:
Previously, if the Literal pattern in Contains, EndsWith and StartsWith had a single quote, the pushdown would fail

With this change, the query is generated correctly 

Shakespeare.py example: df = df.where("word_count > 0 AND word NOT LIKE '%\\'%'") gets correctly generated as 

```
SELECT * FROM ( SELECT ( SUBQUERY_2.SUBQUERY_2_COL_0 ) AS SUBQUERY_3_COL_0 , ( SUM ( SUBQUERY_2.SUBQUERY_2_COL_1 ) ) AS SUBQUERY_3_COL_1 FROM ( SELECT ( SUBQUERY_1.WORD ) AS SUBQUERY_2_COL_0 , ( SUBQUERY_1.WORD_COUNT ) AS SUBQUERY_2_COL_1 FROM ( SELECT * FROM ( SELECT * FROM `bigquery-public-data.samples.shakespeare` AS BQ_CONNECTOR_QUERY_ALIAS ) AS SUBQUERY_0 WHERE ( ( SUBQUERY_0.WORD_COUNT > 0 ) AND NOT ( CONTAINS_SUBSTR ( SUBQUERY_0.WORD , '\'' ) ) ) ) AS SUBQUERY_1 ) AS SUBQUERY_2 GROUP BY SUBQUERY_2.SUBQUERY_2_COL_0 ) AS SUBQUERY_3 ORDER BY ( SUBQUERY_3.SUBQUERY_3_COL_1 ) DESC
```

Screen Shot 2022-09-09 at 2 50 06 PM

### DIFF
--- a/spark-bigquery-pushdown/pushdown_common_src/main/scala/com/google/cloud/spark/bigquery/pushdowns/SparkExpressionConverter.scala
+++ b/spark-bigquery-pushdown/pushdown_common_src/main/scala/com/google/cloud/spark/bigquery/pushdowns/SparkExpressionConverter.scala
@@ -18,6 +18,7 @@ package com.google.cloud.spark.bigquery.pushdowns
 
 import com.google.cloud.bigquery.connector.common.BigQueryPushdownUnsupportedException
 import com.google.cloud.spark.bigquery.pushdowns.SparkBigQueryPushdownUtil.{addAttributeStatement, blockStatement, makeStatement}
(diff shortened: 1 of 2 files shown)


davidrabinowitz (approved):

suryasoma (approved):
LGTM

PR #806, opened 2022-10-17
Add support for EqualNullSafe pushdown

robert3005:
I noticed that this is not implemented but there doesn't seem to be clear reason why it's not as you can translate EqualNullSafe into a sql expression


davidrabinowitz:
Thanks for the PR! Can you please sign the Google CLA so we can accept it?

robert3005:
Signed, anything I need to do to refresh the check?

PR #811, opened 2022-11-02
Accept config for access token provider

gatesn:
I'd like to use a custom AccessTokenProvider but my use-case requires a little bit of configuration that differs for each BQ table that I read. As such, I'd like to be able to pass configuration into the provider via the Spark dataset reader API, e.g. 

```
sparkSession
    .read()
    .format("bigquery")
    .option("gcpAccessTokenProvider", "com.my.token.Provider")
    .option("gcpAccessTokenProviderConfig", "datasetId=foo&tableId=bar")
    .load("foo.bar")
```


gatesn:
Ah, I'd missed spotless. Should be fixed up now

gatesn:
Apologies for pinging again, @suryasoma / @davidrabinowitz does this look good now?

davidrabinowitz:
Closed in favor of PR #913

PR #817, opened 2022-11-10
updating the read me for option clusteredFields, which was confusing

suryasoma:
Fix for the Issue#809


davidrabinowitz (approved):
please update README-template.md as well

PR #818, opened 2022-11-14
Rename BigQueryDirectDataWriterHelper.commit() to finalizeStream()

jphalip:
This is to reflect more accurately what the function does. Note: We can't use just "finalize()" for the name as it's a reserved Java keyword.


davidrabinowitz (approved):

davidrabinowitz (changes requested):
Not sure this is a better description, as the data is not committed. It is committed after `batchCommit()` is called

jphalip:
> Not sure this is a better description, as the data is not committed. It is committed after `batchCommit()` is called

I'm not 100% sure what you're referring to. Are you saying that `finalizeStream()` isn't really a better name for the method? In this case, the data is indeed not committed in BigQuery's parlance so that's why I thought `commit()` was a bit inaccurate. But maybe we're not talking about the same thing?

davidrabinowitz (approved):

PR #823, opened 2022-11-21
convert Readme's datatype tables to standard markdown

JoelAtDeluxe:
This PR addresses one of the TODO's in the readme/readme_template files. Specifically, this revises the data types table to be converted from html syntax to markdown syntax. Markdown doesn't work well with long columns, so for the handful of long cells, I instead placed them as notes below the table.

The styling was converted to markdown: 

* ```<code> -> ` ```
* `<strong> -> **` 
* `<a > -> [label](url)`

But all of the text was otherwise unaltered.

### DIFF
--- a/README-template.md
+++ b/README-template.md
@@ -764,130 +764,34 @@ the options, for example `spark.conf.set("temporaryGcsBucket", "some-bucket")` c
 
 With the exception of `DATETIME` and `TIME` all BigQuery data types directed map into the corresponding Spark SQL data type. Here are all of the mappings:
 
-<!--- TODO(#2): Convert to markdown -->
-<table>
-  <tr valign="top">
-   <td><strong>BigQuery Standard SQL Data Type </strong>
-   </td>
-   <td><strong>Spark SQL</strong>
-<p>
-<strong>Data Type</strong>
-   </td>
-   <td><strong>Notes</strong>
-   </td>
-  </tr>
-  <tr valign="top">
-   <td><strong><code>BOOL</code></strong>
-   </td>
-   <td><strong><code>BooleanType</code></strong>
-   </td>
-   <td>
-   </td>
-  </tr>
-  <tr valign="top">
-   <td><strong><code>INT64</code></strong>
-   </td>
-   <td><strong><code>LongType</code></strong>
-   </td>
-   <td>
-   </td>
-  </tr>
-  <tr valign="top">
-   <td><strong><code>FLOAT64</code></strong>
-   </td>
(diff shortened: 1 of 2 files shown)


README-template.md
```diff
@@ -764,130 +764,34 @@ the options, for example `spark.conf.set("temporaryGcsBucket", "some-bucket")` c
...
-   </td>
-   <td><strong><code>LongType</code></strong>
-   </td>
-   <td>Spark has no TIME type. The generated longs, which indicate <a href="https://avro.apache.org/docs/1.8.0/spec.html#Time+%2528microsecond+precision%2529">microseconds since midnight</a> can be safely cast to TimestampType, but this causes the date to be inferred as the current day. Thus times are left as longs and user can cast if they like.
-<p>
-When casting to Timestamp TIME have the same TimeZone issues as DATETIME
-   </td>
-  </tr>
-</table>
+| **BigQuery Standard SQL Data Type**  | **Spark SQL Data Type**               | **Notes**       |
```
suryasoma, line 767: I feel it is more it is more intuitive to have the notes in the column next to types rather than having it below.
Also, if we are having the notes below, we can delete the column `Notes`, since there it does not have any content in it anyway.

README.md
```diff
@@ -758,130 +758,35 @@ the options, for example `spark.conf.set("temporaryGcsBucket", "some-bucket")` c
...
-   </td>
-   <td><strong><code>LongType</code></strong>
-   </td>
-   <td>Spark has no TIME type. The generated longs, which indicate <a href="https://avro.apache.org/docs/1.8.0/spec.html#Time+%2528microsecond+precision%2529">microseconds since midnight</a> can be safely cast to TimestampType, but this causes the date to be inferred as the current day. Thus times are left as longs and user can cast if they like.
-<p>
-When casting to Timestamp TIME have the same TimeZone issues as DATETIME
-   </td>
-  </tr>
-</table>
+
```
suryasoma, line 761: We do not need to have the same change in README.md, since we overwrite this with README-template.md changes during release.

README-template.md
JoelAtDeluxe: Given the goal of sticking to markdown structures over html, I think it makes more sense to keep the bottom notes and remove the notes column altogether. While I would appreciate a separate column for these notes, the notes themselves are too long to reasonably fit into a column. I'll correct this.

README.md
JoelAtDeluxe: Okay. I can remove this from the readme.

PR #891, opened 2023-02-04
DPP filter only on partition and clustering fields

davidrabinowitz:


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
```diff
@@ -426,6 +427,19 @@ public static Optional<String> getPartitionField(TableInfo tableInfo) {
     return Optional.empty();
```
anoopj, line 427: We can do this as a separate PR, but we would like to include Hive partitioning keys on BigLake tables as well. (Table.ExternalDataConfiguration.HivePartitioningOptions).

anoopj (approved):
LGTM

PR #904, opened 2023-02-16
Fixing premature client closing in certain cases

davidrabinowitz:
This PR reverts some of the code from PR #832


emkornfield (commented):
What are the plans for testing this to make sure we don't have a regression.  Can we still leak connections?

davidrabinowitz:
No, we are actually closing all clients using shutdown hooks, see [here](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/c8ed54702e27a3dd591f265ae30af0072d3e2ada/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientFactory.java#L75)

suryasoma (approved):

PR #906, opened 2023-02-17
Removing row number stats

abmodi:
Row number stats without column stats generate incorrect join reorder plans. Improves performance of TPCDS q19, q85 by around 20x.


vishalkarve15 (approved):
LGTM

PR #910, opened 2023-02-24
Allow spark DSV1 connector to reuse exchange for same table and pushdown filters

vishalkarve15:


vishalkarve15:
DSV1 connector creates duplicate scan relation nodes. Reproduce it using this:

```
val taxiTripsDF =
  (spark.read.format("bigquery")
  .option("table","bigquery-public-data.chicago_taxi_trips.taxi_trips")
  .load())

val taxiTrips2DF =
  (spark.read.format("bigquery")
  .option("table","bigquery-public-data.chicago_taxi_trips.taxi_trips")
  .load())

val taxiTrips3DF =
  (spark.read.format("bigquery")
  .option("table","bigquery-public-data.chicago_taxi_trips.taxi_trips")
  .load())
val taxiTrips4DF =
  (spark.read.format("bigquery")
  .option("table","bigquery-public-data.chicago_taxi_trips.taxi_trips")
  .load())

taxiTripsDF.createOrReplaceTempView("taxi_trips")
taxiTrips2DF.createOrReplaceTempView("taxi_trips2")
taxiTrips3DF.createOrReplaceTempView("taxi_trips3")
taxiTrips4DF.createOrReplaceTempView("taxi_trips4")

val answerDF = spark.sql("""
  SELECT
 (SELECT max(a.unique_key) FROM taxi_trips AS a JOIN taxi_trips2 AS b ON b.unique_key = a.unique_key),
 a.unique_key
FROM taxi_trips3 AS a
JOIN taxi_trips4 AS b ON b.unique_key = a.unique_key
LIMIT 100
  """)

val res = answerDF.collect()
```
Reproduced plan: [image link](https://user-images.githubusercontent.com/25999807/221107241-6b195fb9-1f9b-41c9-b8bf-11d1948c876e.jpg)

The fix is to implement equals, hashcode for the datasource v1 relation.

Fixed plan: [image link](https://user-images.githubusercontent.com/25999807/221107266-5d95d1dc-897c-4885-8198-f40233f260c7.jpg)

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.java
```diff
@@ -63,6 +64,8 @@ public class DirectBigQueryRelation extends BigQueryRelation
   private final TableDefinition defaultTableDefinition;
   private final BigQueryRDDFactory bigQueryRDDFactory;
 
+  private Filter[] pushedFilters = new Filter[] {};
```
davidrabinowitz, line 67: It seems that the only usage for this field is in equals() and hashcode(), and in both cases we compare (or rely on) the result of `getCompiledFilter(pushedFilters)`. Please save the compiled filter instead - as we calculate it anyway, there's no reason to do it twice.

davidrabinowitz (changes requested):

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/direct/DirectBigQueryRelation.java
vishalkarve15: Thanks for the review, addressed.

davidrabinowitz (approved):

PR #921, opened 2023-03-14
Remove one level of indirection in ArrowSchemaConverter.

emkornfield:
- Make ArrowSchemaConverter Generic
- Changes all accessors to now extend ArrowSchemaConverter
- Instantiates new classes via static factory method instead of direct constructor.
- Moves null checks to each concrete implementation.  While I believe JVM should be able to optimize based on the parent method, I'm not 100% sure and this way JVM should be able to avoid megamorphic dispatch for sure (also eliminate old work around from a [bug in Arrow 0.9](https://github.com/apache/spark/commit/e50a37a587fd7db48b59e37f234f8c74bd6707f7) for checks.
- Move child fields to StructAccessor (hopefully keeps more common classes smaller).


emkornfield:
Closed https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/920

emkornfield:
@davidrabinowitz I thought I replied here, but I think changes are probably ready for review (I'm not sure what to do about codecov)

davidrabinowitz (approved):

PR #955, opened 2023-04-24
[BREAKING CHANGE] Numeric/Decimal updates

davidrabinowitz:
This PR contains the following fixes:

* BigNumeric will be converted to Decimal, not the UDT that was used before. Notice that certain schemas won't be supported as BigNumeric can be wider than Spark's Decimal (BigNumeric's maximal precision is 76 whereas Spark is limited to 38)
* All numeric will now reflect the same Precision and scale as the other system


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SchemaConverters.java
```diff
@@ -337,9 +329,31 @@ private DataType getStandardDataType(Field field) {
...
+      return DataTypes.createDecimalType(precision, scale);
     } else if (LegacySQLTypeName.BIGNUMERIC.equals(field.getType())) {
-      return BigQueryDataTypes.BigNumericType;
+      int precision =
+          Optional.ofNullable(field.getPrecision())
+              .map(Long::intValue)
+              .orElse(BigQueryUtil.DEFAULT_BIG_NUMERIC_PRECISION);
+      if (precision > DecimalType.MAX_PRECISION()) {
+        throw new IllegalArgumentException(
+            String.format(
```
suryasoma, line 348: Can we add a test for the Exception case

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
```diff
@@ -921,6 +923,45 @@ public void testWriteToCmekManagedTable() throws Exception {
         .isEqualTo(destinationTableKmsKeyName);
   }
 
+  @Test
+  public void testWriteNumericsToWiderFields() throws Exception {
```
suryasoma, line 927: can we add an exception case as well?

davidrabinowitz: Why do we need it in the integration test, if the check is done locally?

bigquery-connector-common/src/test/java/com/google/cloud/bigquery/connector/common/BigQueryUtilTest.java
```diff
@@ -251,6 +253,7 @@ public void testFieldWritableMaxLength() {
   }
 
   @Test
+  @Ignore
```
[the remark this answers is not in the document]
davidrabinowitz: Removed the tests, the logic is tested in the new tests in the class

suryasoma (approved):

PR #957, opened 2023-04-26
Direct write with subset field list support fix

suryasoma:


bigquery-connector-common/src/test/java/com/google/cloud/bigquery/connector/common/BigQueryUtilTest.java
```diff
@@ -303,9 +309,48 @@ public void testSchemaWritableWithDisableNullableFieldCheck() {
...
+
+    assertThat(BigQueryUtil.schemaWritable(s1, s2, false, true)).isTrue();
+    assertThat(BigQueryUtil.schemaWritable(s1, s3, false, true)).isFalse();
+    assertThat(BigQueryUtil.schemaWritable(s3, s2, false, true)).isTrue();
+  }
+
+  @Test
+  public void testIsModeWritable() {
+    assertThat(BigQueryUtil.isModeWritable(Mode.NULLABLE, Mode.NULLABLE)).isTrue();
+    assertThat(BigQueryUtil.isModeWritable(Mode.NULLABLE, Mode.REQUIRED)).isTrue();
```
davidrabinowitz, line 346: Why writing nullable to required is ok? it may contain `null`s

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
```diff
@@ -322,6 +332,18 @@ static boolean fieldWritable(
                 && sourceField.getPrecision() <= destinationField.getPrecision()));
   }
 
+  @VisibleForTesting
+  static boolean isModeWritable(Field.Mode sourceMode, Field.Mode destinationMode) {
+    switch (destinationMode) {
+      case REPEATED:
+        return sourceMode == Mode.REPEATED;
+      case REQUIRED:
```
davidrabinowitz, line 340: Can you write nullable into required?

davidrabinowitz (changes requested):

bigquery-connector-common/src/test/java/com/google/cloud/bigquery/connector/common/BigQueryUtilTest.java
suryasoma: So here we are only checking the schema not the actual value. 
So here is the case if the destination field is `REQUIRED` and the source is `NULLABLE`, but if the source contains non-null value, it should be written. The check here is to see if schema is compatible not the values.

davidrabinowitz (changes requested):
Please add integration tests checking all the cases

davidrabinowitz (approved):

PR #965, opened 2023-05-05
changing the hash from BigQueryConfig to BigQueryClientFactoryConfig

suryasoma:
BigQueryConfig doesn't overwrite hashCode() and equals(), causing to create a client for every request.

### DIFF
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientFactory.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientFactory.java
@@ -113,10 +113,10 @@ public class BigQueryClientFactory implements Serializable {
       return Objects.hashCode(
           BigQueryUtil.getCredentialsByteArray(credentials),
           headerProvider,
-          new BigQueryClientFactoryConfig(bqConfig));
+          bqConfig.clientCreationHashCode());
     }
 
-    return Objects.hashCode(credentials, headerProvider, new BigQueryClientFactoryConfig(bqConfig));
+    return Objects.hashCode(credentials, headerProvider, bqConfig.clientCreationHashCode());
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientFactory.java
```diff
@@ -111,10 +111,12 @@ public int hashCode() {
     // ExternalAccountCredentials first and then compare their hashCodes.
     if (credentials instanceof ExternalAccountCredentials) {
       return Objects.hashCode(
-          BigQueryUtil.getCredentialsByteArray(credentials), headerProvider, bqConfig);
+          BigQueryUtil.getCredentialsByteArray(credentials),
+          headerProvider,
+          new BigQueryClientFactoryConfig(bqConfig));
```
davidrabinowitz, line 116: To avoid duplication, please move this conversion to the constructor (line 71)

davidrabinowitz (changes requested):

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientFactory.java
suryasoma: Ack

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryConfig.java
```diff
@@ -65,4 +66,40 @@ public interface BigQueryConfig {
...
+        getAccessTokenProviderConfig(),
+        getCredentialsKey(),
+        getAccessToken(),
+        getCredentialsFile(),
+        getBigQueryHttpEndpoint(),
+        getFlowControlWindowBytes(),
+        getBigQueryStorageGrpcEndpoint(),
+        getCreateReadSessionTimeoutInSeconds(),
+        getBigQueryProxyConfig(),
+        getParentProjectId(),
```
davidrabinowitz, line 82: Why do we need the parent project ID ? it is not used by the factory

davidrabinowitz (changes requested):

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryConfig.java
suryasoma: It is used while creating `HeaderProvider` https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/9e31d999f8c7e80121b48a32a24d0d28c64189c7/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/HttpUtil.java#L30
Added just to ensure all the dependencies are equal.

davidrabinowitz (approved):

PR #979, opened 2023-05-24
Create spark table from BQ table fix

suryasoma:
fix to create spark table from BQ table, created in the same flow


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/write/BigQueryInsertableRelationBase.java
```diff
@@ -71,7 +73,7 @@ public boolean isEmpty() {
 
   /** Returns the number of rows in the table. If the table does not exist return None */
   private Optional<BigInteger> numberOfRows() {
-    return table.map(TableInfo::getNumRows);
+    return Optional.of(table.get().getNumRows());
```
vishalkarve15, line 76: `table.get()` can return null. We should add a check and return `Optional.empty()` if that's the case.

davidrabinowitz: This is why we are using map() - if the table is empty it will return empty. This suggestion will throw NoSuchElementException

Issue #144, opened 2020-03-24
Unable to insert data into Time Column from spark scala.


charan478:
I am trying to insert data into bigquery table with Time column from spark . I am currently trying with 0.13.1-beta, I have also tried with 0.11.1-beta which gave me same error. I am aware that Spark sql have no TimeType and when reading data from bigquery to spark it converts that into long type. I am trying to insert column of type long into field of Time type in bigquery and failing with error that the column is integer. I am confused why the type is infered as integer when the type of scrum_time column is long in spark. I tried intermediate type parquet and orc no diffrence 

Snippet I am using to push to Bigquery in spark is

```
batchDf.printSchema()
batchDf.show(false)
tables.foreach(table=>batchDf.write.mode(SaveMode.Append).format("com.google.cloud.spark.bigquery").option("temporaryGcsBucket", "sparkoperatorexample").option("table", table).save()
```
The schema for the row I am trying to insert into bigquery is 
```20/03/24 07:34:22 INFO CodeGenerator: Code generated in 415.620557 ms
root
 |-- name: string (nullable = true)
 |-- age: integer (nullable = true)
 |-- birthday: timestamp (nullable = true)
 |-- employed_date: date (nullable = true)
 |-- scrum_time: long (nullable = true)

Few Sample data is as follows
20/03/24 07:34:41 INFO DAGScheduler: Job 0 finished: start at ConfluentKafkaAvroReader.scala:84, took 17.865636 s
+------+---+-------------------+-------------+----------+
|name  |age|birthday           |employed_date|scrum_time|
+------+---+-------------------+-------------+----------+
|ching |22 |1991-06-28 18:50:20|2019-09-03   |39650099  |
|charan|28 |1991-06-28 13:17:00|2019-09-03   |39650099  |
|ching |22 |1991-06-28 18:50:20|2019-09-03   |39650099  |
|charan|28 |1991-06-28 13:17:00|2019-09-03   |39650099  |
|ching |22 |1991-06-28 18:50:20|2019-09-03   |39650099  |
|charan|28 |1991-06-28 13:17:00|2019-09-03   |39650099  |
|ching |22 |1991-06-28 18:50:20|2019-09-03   |39650099  |
|charan|28 |1991-06-28 13:17:00|2019-09-03   |39650099  |
+------+---+-------------------+-------------+----------+
```

Table Schema 
```
Field name	Type		Mode		Description
name		STRING		NULLABLE	employee name
[...]

PR #1017, opened 2023-07-14
Issue #144: allow writing Spark String to BQ TIME type

vishalkarve15:
Currently, we allow reading BQ's TIME datatype into Spark's long datatype. However, there's no way to write into BQ's TIME datatype.
This change adds that functionality through String.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Next
 
+* Issue #144: allow writing Spark String to BQ TIME type
 * PR #1038: Logical plan now shows the BigQuery table of DirectBigQueryRelation. Thanks @idc101 !
 
 ## 0.32.2 - 2023-08-07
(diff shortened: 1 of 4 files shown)


vishalkarve15:
@davidrabinowitz do you think we should also allow writing strings to TIME? This will allow users to write strings in `[H]H:[M]M:[S]S[.DDDDDD|.F]` format to TIME.

davidrabinowitz (changes requested):
What happens when a user reads a BQ table and then tries to write the DataFrame to another table? 

Also, please add a section to `README-template.md` explaining this option to the users

vishalkarve15:
> What happens when a user reads a BQ table and then tries to write the DataFrame to another table?
> 
> Also, please add a section to `README-template.md` explaining this option to the users

Reading a BQ table with a TIME column will create a dataframe of Long datatype.
We cannot write Long datatype back into a BQ table as TIME.

davidrabinowitz (approved):

PR #1025, opened 2023-07-19
Handle Java 8 types for dates and timestamps when compiling filters

tom-s-powell:
The `spark.sql.datetime.java8API.enabled` configuration as described in https://spark.apache.org/docs/3.0.0/configuration.html is not currently handled (when set to `true`) when compiling filters. Date or timestamp literals will be `java.time.LocalDate` or `java.time.Instant` respectively and thus not properly escaped by https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/master/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkFilterUtils.java#L318 resulting in query failures due to invalid syntax.

### DIFF
--- a/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkFilterUtils.java
+++ b/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkFilterUtils.java
@@ -23,7 +23,6 @@ import java.sql.Date;
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.LocalDate;
-import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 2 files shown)


tom-s-powell:
Fixed up failing test.

davidrabinowitz:
> @davidrabinowitz thoughts on the test coverage issue?

This delta is unclear, and is concentrated at an unreleased code, so it is safe to ignore for now.

Thanks for the PR!

davidrabinowitz (approved):

tom-s-powell:
@davidrabinowitz is there an ETA on next release?

davidrabinowitz:
This week 🤞

PR #1032, opened 2023-07-31
Reverting changes for overwrite mode to use MERGE, to prevent loosing policy tags

suryasoma:
As part of PR https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/1007, write has been changed from `MERGE` to `SELECT` with respective WriteDisposition. But in overwrite mode with WriteDisposition as `WRITE_TRUNCATE` the policyTags are getting replaced. So reverting this change from #1007

### DIFF
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
@@ -236,7 +236,22 @@ public class BigQueryClient {
    */
   public Job overwriteDestinationWithTemporary(
       TableId temporaryTableId, TableId destinationTableId) {
-    return copyData(temporaryTableId, destinationTableId, JobInfo.WriteDisposition.WRITE_TRUNCATE);
+    String queryFormat =
+        "MERGE `%s`\n"
+            + "USING (SELECT * FROM `%s`)\n"
+            + "ON FALSE\n"
+            + "WHEN NOT MATCHED THEN INSERT ROW\n"
+            + "WHEN NOT MATCHED BY SOURCE THEN DELETE";
+
+    QueryJobConfiguration queryConfig =
+        jobConfigurationFactory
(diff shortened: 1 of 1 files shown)


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -213,10 +213,11 @@ private Job copyData(
       TableId destinationTableId,
       JobInfo.WriteDisposition writeDisposition) {
     String queryFormat = "SELECT * FROM `%s`";
+    String temporaryTableName = fullTableName(sourceTableId);
```
agrawal-siddharth, line 216: Changes to the copyData() function are not needed.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -252,11 +269,6 @@ public Job appendDestinationWithTemporary(TableId temporaryTableId, TableId dest
     return copyData(temporaryTableId, destinationTableId, JobInfo.WriteDisposition.WRITE_APPEND);
   }
 
-  String sqlFromFormat(String queryFormat, TableId temporaryTableId) {
```
agrawal-siddharth, line 255: Similarly, removal of this function is not needed.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
suryasoma: it is the same method which has been shifted into here since we will have two `sqlFromFormat` methods one for `SELECT` and the other for `MERGE`.  Since the format methods are closely tied to the queries, I have dropped them and made changes to the caller itself.

agrawal-siddharth (approved):

PR #1033, opened 2023-07-31
Fixing numeric to decimal conversion

davidrabinowitz:


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SchemaConverters.java
```diff
@@ -408,6 +399,35 @@ private DataType getStandardDataType(Field field) {
     }
   }
 
+  @VisibleForTesting
+  static DecimalType createDecimalTypeFromNumericField(Field field) {
```
suryasoma, line 403: can we create the same check if the SQL type is `BIGNUMERIC` as well? https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/741bdb33bf5c06fb3fad9defffa65ea1470d2690/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SchemaConverters.java#L375

suryasoma (approved):

PR #1055, opened 2023-08-31
async appendRows response validation

vishalkarve15:


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
```diff
@@ -253,6 +259,13 @@ public long finalizeStream() throws IOException {
     if (this.protoRows.getSerializedRowsCount() != 0) {
       sendAppendRowsRequest();
     }
+    try {
+      ApiFutures.allAsList(appendRowsFutures).get();
```
davidrabinowitz, line 263: There's no validation on the result. What happens if one of the validateAppendRowsResponse() calls throws exception? It is done in another thread

vishalkarve15: It is done on line 206-211. These are validated futures.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
```diff
@@ -253,6 +259,13 @@ public long finalizeStream() throws IOException {
...
+    } catch (InterruptedException | ExecutionException e) {
+      throw new BigQueryConnectorException("Could not retrieve AppendRowsResponse", e);
+    } finally {
+      appendRowsExecutor.shutdown();
+    }
```
agrawal-siddharth, line 268: A consequence of checking the async results here in finalizeStream() is that all (unless throttled by the write client library) rows will get sent regardless of an earlier failure. This could cause greater stress on the server. For example, if a request resulted in a resource-exceeded quota error, the additional rows would be sent anyway. One way to mitigate this is to check the status of any outstanding responses that may have already completed, at the beginning of sendAppendRowsRequest(). If a failure is detected at that time, we can throw an exception and cause an abort right away.

agrawal-siddharth (changes requested):

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
```diff
@@ -63,6 +68,9 @@ public class BigQueryDirectDataWriterHelper {
   private long appendRequestSizeBytes = 0; // number of bytes waiting for the next append request
   private long writeStreamRowCount = 0; // total offset / rows of the current write-stream
 
+  private final ExecutorService appendRowsExecutor = Executors.newSingleThreadExecutor();
```
arunkumarchacko, line 71: Do you think making this static and having a common threadpool is beneficial?

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
```diff
@@ -253,6 +259,13 @@ public long finalizeStream() throws IOException {
     if (this.protoRows.getSerializedRowsCount() != 0) {
       sendAppendRowsRequest();
     }
+    try {
+      ApiFutures.allAsList(appendRowsFutures).get();
+    } catch (InterruptedException | ExecutionException e) {
+      throw new BigQueryConnectorException("Could not retrieve AppendRowsResponse", e);
```
arunkumarchacko, line 265: Should we log and throw? Also include some debug information along with the message. eg: tablePath

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
```diff
@@ -253,6 +259,13 @@ public long finalizeStream() throws IOException {
...
+    } finally {
+      appendRowsExecutor.shutdown();
```
arunkumarchacko, line 267: should we add a check at other public methods to assert that no operation is done after the executor is shutdown?

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
```diff
@@ -195,8 +203,12 @@ private void sendAppendRowsRequest() throws IOException {
 
     ApiFuture<AppendRowsResponse> appendRowsResponseApiFuture =
         streamWriter.append(protoRows.build(), offset);
-    validateAppendRowsResponse(appendRowsResponseApiFuture, offset);
-
+    ApiFuture<AppendRowsResponse> validatedAppendRowsFuture =
+        ApiFutures.transformAsync(
+            appendRowsResponseApiFuture,
+            appendRowsResponse -> validateAppendRowsResponse(appendRowsResponse, offset),
```
arunkumarchacko, line 209: Is it ok if there is a race b/w rows? For e.g. a row1 is inserted by client at t1 and row2 inserted at t2 (t2 > t1), but row2 got inserted in the table before row1.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
agrawal-siddharth: This change has been made in https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/1057

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
agrawal-siddharth: This change has been made in https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/1057

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
agrawal-siddharth: The executor will throw an exception if it is used after being shutdown.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
agrawal-siddharth: Such a race should not occur. This is because the requests are being inserted on a single thread. With bidi streaming we are guaranteed that requests will be sent in order, and corresponding responses will arrive in order.

PR #1058, opened 2023-09-08
View names will appear in query plan instead of the materialized table

davidrabinowitz:


isha97 (approved):

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
```diff
@@ -594,4 +594,11 @@ static Field adjustField(Field field, @Nullable Field existingField) {
     // no adjustment
     return field;
   }
+
+  public static String prepareQueryForLog(String query, int maxLength) {
+    String noNewLinesQuery = query.replace("\n", "");
+    return noNewLinesQuery.length() > maxLength
+        ? noNewLinesQuery.substring(0, maxLength) + /* ellipsis */ '\u2026'
```
arunkumarchacko, line 601: can consider logging the truncated part length.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
```diff
@@ -594,4 +594,11 @@ static Field adjustField(Field field, @Nullable Field existingField) {
     // no adjustment
     return field;
   }
+
+  public static String prepareQueryForLog(String query, int maxLength) {
+    String noNewLinesQuery = query.replace("\n", "");
```
vishalkarve15, line 599: A suggestion is to replace "\n" with "\\n". This makes it more readable if there are comments.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
davidrabinowitz: The full query is in the log. We put the first 50 characters, and it is just for the query plan

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
davidrabinowitz: Done

PR #1061, opened 2023-09-12
Handle NPE case when reading BQ table with NUMERIC fields

hayssams:
Fix Null pointer exception when updating an existing bigquery table with NUMERIC attributes without scale and precision explicitly set

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Next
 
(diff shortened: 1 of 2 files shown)


vishalkarve15:
This happens when "readDataFormat" option is set to "avro".

vishalkarve15 (approved):

vishalkarve15:
Thank you for your contribution, @hayssams !

PR #1082, opened 2023-10-05
[WIP] request snappy compression of ReadRowsResponse and decompress compressed messages

yqiu2:
this does not build because the new proto fields in the TableReadOptions and the ReadRowsResponse do not exist (yet).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ AdAdHocITSuite.scala
 
 # Mac
 .DS_Store
+
+# Visual Studio IDE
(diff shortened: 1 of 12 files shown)


yqiu2:
How can I populate an unknown proto field, response_compression_codec, and how can I read unknown proto field uncompressed_byte_size?   

alternatively,

can I generate a custom private Java Client library containing new proto fields, and make this connector not resolve  dependencies to maven, but to my custom private Java Client library?  Will this work?

PR #1094, opened 2023-10-18
CVE-2023-5072: Upgrading the org.json:json dependency

davidrabinowitz:


isha97:
There are additional code duplication removal commits which should not be part of this PR.

isha97 (approved):

PR #1096, opened 2023-10-18
Support Dynamic Partition Overwrite [DIRECT write, Range Partitioning]

isha97:


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -254,6 +255,13 @@ public Job overwriteDestinationWithTemporaryDynamicPartitons(
             getQueryForTimePartitionedTable(
                 destinationTableName, temporaryTableName, sdt, timePartitioning);
       }
+
+      RangePartitioning rangePartitioning = sdt.getRangePartitioning();
```
davidrabinowitz, line 259: Should be in an `else` clause

davidrabinowitz (changes requested):

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
isha97: Done

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
```diff
@@ -639,4 +633,68 @@ static String getQueryForTimePartitionedTable(
...
+      StandardTableDefinition destinationDefinition,
+      RangePartitioning rangePartitioning) {
+
+    long start = rangePartitioning.getRange().getStart();
+    long end = rangePartitioning.getRange().getEnd();
+    long interval = rangePartitioning.getRange().getInterval();
+
+    String partitionField = rangePartitioning.getField();
+    String extractedPartitioned =
+        "IF(%s.%s > %s, 0, RANGE_BUCKET(%s.%s, GENERATE_ARRAY(%s, %s, %s)))";
```
vishalkarve15, line 669: Should this be >= ? 
The end is exclusive and not inclusive.
https://cloud.google.com/bigquery/docs/partitioned-tables#integer_range

vishalkarve15, line 669: Do we need a similar condition to check if it's < start?

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
```diff
@@ -1797,6 +1797,94 @@ public void testOverwriteDynamicPartition_noTimePartitioning() {
...
+
+    assertThat(row4.getLong(row4.fieldIndex(orderId))).isEqualTo(85);
+    assertThat(row4.getLong(row4.fieldIndex(orderCount))).isEqualTo(3000);
+
+    assertThat(row5.getLong(row5.fieldIndex(orderId))).isEqualTo(90);
+    assertThat(row5.getLong(row5.fieldIndex(orderCount))).isEqualTo(3050);
+  }
+
+  @Test
+  public void testOverwriteDynamicPartition_rangePartitionedOutsideRange() {
```
vishalkarve15, line 1853: Let's test outside the range from below as well.

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
```diff
@@ -1797,6 +1797,94 @@ public void testOverwriteDynamicPartition_noTimePartitioning() {
         .isEqualTo(Timestamp.valueOf("2023-09-30 12:00:00"));
   }
 
+  @Test
+  public void testOverwriteDynamicPartition_rangePartitioned() {
```
vishalkarve15, line 1801: According to https://cloud.google.com/bigquery/docs/partitioned-tables#integer_range, values outside the range go into a partition called `__UNPARTITIONED__` and null values go into a partition called `__NULL__`.
Can we do that as well? If it's too complicated, let's call out in the README that the connector doesn't support this behaviour.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
isha97: Good catch. Fixed!

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
isha97: Done

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryUtil.java
isha97: No by default if it's < start it puts the row in bucket 0 so all out of range will be in bucket 0

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
isha97: We are supporting both

davidrabinowitz (approved):

PR #1097, opened 2023-10-18
Upgraded dependencies

davidrabinowitz:


CHANGES.md
```diff
@@ -2,6 +2,12 @@
 
 ## Next
 
+* BigQuery API has been upgraded to version 2.33.2
+* BigQuery Storage API has been upgraded to version 2.44.0
+* GAX has been upgraded to version 2.35.0
+* gRPC has been upgraded to version 1.58.0
+* Protocol Buffers has been upgraded to version 3.24.4
+
```
isha97, line 10: Do we need to document all version upgrades? I don;t see the dataproc, gcs version upgrades here

davidrabinowitz: We've been doing that for the past few years. It's also the common practice for our dependencies

isha97 (approved):

PR #1104, opened 2023-10-24
Fixed nested field predicate pushdown

davidrabinowitz:


isha97 (approved):

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkFilterUtils.java
```diff
@@ -338,6 +338,8 @@ static String escape(String value) {
   }
 
   static String quote(String value) {
-    return "`" + value + "`";
+    return value.contains(".")
+        ? Arrays.stream(value.split("\\.")).collect(Collectors.joining("`.`", "`", "`"))
```
satybald, line 342: could we add a comment to describe better what's going on here?

davidrabinowitz: Added comment, thanks for the suggestion

PR #1118, opened 2023-11-04
Improve read session caching key

vishalkarve15:
We use pushed filters indirectly as a key in read session cache. However, it misses caching opportunities for duplicated filters and out of order filters.
This fixes that.


vishalkarve15:
Saw ~100+s improvement in TPCDS benchmarks.

anoopj (approved):
LGTM!

PR #1136, opened 2023-12-01
Consider projections for biglake stats

vishalkarve15:
Biglake tables do not take projections into account when estimating size. This makes Spark work with an overestimated size causing missed opportunities for Broadcast.


anoopj (approved):
LGTM

PR #1143, opened 2023-12-12
Enable async write for default stream.

agrawal-siddharth:


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
```diff
@@ -252,13 +252,7 @@ private void checkForFailedResponse(boolean waitForResponse) {
...
    */
   private void sendAppendRowsRequest() throws IOException {
-    boolean waitForResponse = false;
-    if (writeAtLeastOnce) {
-      // FUTURE: Enable async writes with default stream once client retries are supported.
-      // When using default stream we may encounter retry-able server errors when under high load.
-      waitForResponse = true; // disable async writes
-    }
-    checkForFailedResponse(waitForResponse);
+    checkForFailedResponse(false /*waitForResponse*/);
```
vishalkarve15, line 255: should we make changes in finalizeStream as well?
```
    if (!this.writeAtLeastOnce) {
      waitBeforeFinalization();
```

agrawal-siddharth: No. No changes are needed in finalizeStream(). There, we first call checkForFailedResponse(true) to wait for all outstanding responses to be received, to ensure there were no errors. Then, if we are using the default stream there is no need to call Finalize() on it.

vishalkarve15 (approved):

Issue #1116, opened 2023-10-31
BigQuery write fails with MessageSize is too large


eric-pa:
Hi,

I'm trying to write a dataframe to a BigQuery table which seems to fails due to the Storage Write API limit of 10 MB per request.

### Code
The code looks as follows:
```py
df.write \
  .mode("overwrite") \
  .format("bigquery") \
  .option("writeMethod", "direct") \
  .option("writeDisposition", 'WRITE_TRUNCATE') \
  .save("xxx.xxx.xxx")
```

### Cluster configuration:
- serverless batch job
- spark-bigquery-with-dependencies_2.12-0.33.0.jar

### Error
Spark output:
```
Using the default container image
Waiting for container log creation
PYSPARK_PYTHON=/opt/dataproc/conda/bin/python
Generating /home/spark/.pip/pip.conf
Configuring index-url as 
Configuring target as /mnt/dataproc/python/site-packages
JAVA_HOME=/usr/lib/jvm/temurin-11-jdk-amd64
SPARK_EXTRA_CLASSPATH=
:: loading settings :: file = /etc/spark/conf/ivysettings.xml
23/10/31 16:42:38 WARN BigQueryDataSourceWriterInsertableRelation: It seems that 1 out of 0 partitions have failed, aborting
23/10/31 16:42:38 WARN BigQueryDirectDataSourceWriterContext: BigQuery Data Source writer xxxxxxx-xxx aborted

```

Logs:
```
com.google.cloud.bigquery.connector.common.BigQueryConnectorException: Could not retrieve AppendRowsResponse
    at com.google.cloud.bigquery.connector.common.BigQueryDirectDataWriterHelper.validateAppendRowsResponse(BigQueryDirectDataWriterHelper.java:212)
    at com.google.cloud.bigquery.connector.common.BigQueryDirectDataWriterHelper.sendAppendRowsRequest(BigQueryDirectDataWriterHelper.java:188)
    at com.google.cloud.bigquery.connector.common.BigQueryDirectDataWriterHelper.addRow(BigQueryDirectDataWriterHelper.java:169)
    at com.google.cloud.spark.bigquery.write.context.BigQueryDirectDataWriterContext.write(BigQueryDirectDataWriterContext.java:86)
    at com.google.cloud.spark.bigquery.write.context.BigQueryDirectDataWriterContext.write(BigQueryDirectDataWriterContext.java:35)
[...]

PR #1159, opened 2024-01-06
Issue #1116: BigQuery write fails with MessageSize is too large

agrawal-siddharth:


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
```diff
@@ -176,8 +176,10 @@ private StreamWriter createStreamWriter(String writeStreamName) {
    */
   public void addRow(ByteString message) throws IOException {
     int messageSize = message.size();
```
vishalkarve15, line 178: Could we just add 2 here? In messageSize?

vishalkarve15 (approved):

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryDirectDataWriterHelper.java
agrawal-siddharth: Done

PR #1168, opened 2024-01-12
Mark Spark33BigQueryPushdown as supporting Spark 3.4 and 3.5

tom-s-powell:
I see in https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/1115/files#diff-87a38904e32856997cfefcedf7cdda864282b53af31200a1a22da4148a6399d0R40 that Spark 3.5 (and Spark 3.4) libs are relying on `spark-3.3-bigquery-pushdown` libs.

However, `Spark33BigQueryPushdown#supportsSparkVersion` only declares it supports 3.3. This means `BigQueryConnectorUtils#disablePushdownSession` does not work on Spark 3.4 or 3.5.

I assumed this was a reasonable fix given there are not 3.4 and 3.5 specific "bigquery-pushdown" libs.


vishalkarve15:
This would need a bigger effort. We need separate directories for `spark-3.4-bigquery-pushdown_2.12`, `spark-3.4-bigquery-pushdown_2.13`, `spark-3.5-bigquery-pushdown_2.12`, `spark-3.5-bigquery-pushdown_2.13` inside the `spark-bigquery-pushdown` to at least cover integration tests.

sid-habu:
@vishalkarve15 What is the alternative that we can use to enable Pushdown in Spark 3.4.2?

PR #1171, opened 2024-01-16
fix read, write issues with Timestamp

vishalkarve15:
This PR fixes 2 things:
- an overflow error in SparkBigQueryUtil that caused off-by-one errors
- error while reading a BQ Timestamp field in non-GMT local timezone


vishalkarve15:
This only works for Spark 3 DSv2. Also works for Scala 2.13 DSv1 since it compiles against Spark 3.2

isha97 (approved):

PR #1197, opened 2024-03-08
Displaying the read session id on spark UI

isha97:


vishalkarve15:
Please verify that description() is not called before build().

isha97:
Verified that the description() is called after build(). Also, added a wrapper to make sure that we never build the read session in description()

spark-bigquery-dsv2/spark-bigquery-dsv2-common/src/main/java/com/google/cloud/spark/bigquery/v2/context/BigQueryDataSourceReaderContext.java
```diff
@@ -531,4 +560,11 @@ private long getRowSize(Collection<StructField> fields) {
             .map(structField -> (long) structField.dataType().defaultSize())
             .reduce(0L, Long::sum);
   }
+
+  public String getReadSessionId() {
+    if (readSessionResponse.isCalculated()) {
```
vishalkarve15, line 565: Why not use `isBuilt` to check this?

isha97: Didn't notice that. Fixed.

isha97: Actually, we can't rely on `isBuilt` because it's set to true just when the create is submitted to the executor service so it might not be built.

davidrabinowitz (changes requested):

spark-bigquery-dsv2/spark-bigquery-dsv2-common/src/main/java/com/google/cloud/spark/bigquery/v2/context/BigQueryDataSourceReaderContext.java
vishalkarve15: That's ok. We can call readSession.get after `isBuilt` is true. It's thread safe.

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
```diff
@@ -2615,6 +2615,31 @@ public void testTimestampNTZIndirectWriteToBigQueryParquetFormat() throws Interr
...
+    Dataset<Row> readDF =
+        spark
+            .read()
+            .format("bigquery")
+            .option("dataset", testDataset.toString())
+            .option("table", testTable)
+            .load();
+
+    assertThat(readDF.count()).isAtLeast(1);
+    assertEquals(
```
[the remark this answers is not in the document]
isha97: Fixed

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
```diff
@@ -2615,6 +2615,31 @@ public void testTimestampNTZIndirectWriteToBigQueryParquetFormat() throws Interr
...
+  @Test
+  public void testTableDescriptionRemainsUnchanged() {
+    IntegrationTestUtils.runQuery(
+        String.format("CREATE TABLE `%s.%s` (name STRING, age INT64)", testDataset, testTable));
+    String initialDescription = bq.getTable(testDataset.toString(), testTable).getDescription();
+    Dataset<Row> df =
+        spark.createDataFrame(
+            Arrays.asList(RowFactory.create("foo", 10), RowFactory.create("bar", 20)),
+            new StructType().add("name", DataTypes.StringType).add("age", DataTypes.IntegerType));
+    writeToBigQueryAvroFormat(df, SaveMode.Append, "True");
```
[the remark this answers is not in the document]
isha97: The `writeToBigQueryAvroFormat` calls `writeToBigQuery` which uses the write method specified by the class of integration test. The format is also set in the config which would be used if `writeMethod` is `indirect`

spark-bigquery-dsv2/spark-bigquery-dsv2-common/src/main/java/com/google/cloud/spark/bigquery/v2/context/BigQueryDataSourceReaderContext.java
isha97: It doesn't work as expected in integration tests and try to compute the read session twice. Example test : `testPartition_Daily`

PR #1221, opened 2024-05-01
Add config option for snapshot time

tom-s-powell:


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
```diff
@@ -218,6 +218,7 @@ public static WriteMethod from(@Nullable String writeMethod) {
       com.google.common.base.Optional.absent();
   private boolean enableReadSessionCaching = true;
   private long readSessionCacheDurationMins = 5L;
+  private OptionalLong snapshotTimeMillis = OptionalLong.empty();
```
davidrabinowitz, line 221: SparkBigQueryConfig is a Serializable class, and therefore we need to keep all of it's members  Serializable as well. This is why all `Optional`s are the guava version and not Java's. Please replace this OptionalLong with `com.google.common.base.Optional<Long>`? The conversion to `OptionalLong` should be done in `getSnapshotTimeMillis()`

davidrabinowitz (changes requested):
Thanks for the contribution! Please see the comment below.

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
tom-s-powell: Have switched to have similar behaviour to `partitionExpirationMs`

davidrabinowitz (approved):

PR #1222, opened 2024-05-01
add option to request lz4 compressed ReadRowsResponse

yqiu2:
This adds the ability to request compressed ReadRowsResponses.
If the server decides to give back a ReadRowsResponse, we use the specified compression algorithm (LZ4) to decompress the results. 
This also adds two integration tests to test requesting compressed results for both arrow and avro.

### DIFF
--- a/bigquery-connector-common/pom.xml
+++ b/bigquery-connector-common/pom.xml
@@ -28,6 +28,11 @@
       <artifactId>proto-google-cloud-bigquerystorage-v1</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.lz4</groupId>
+      <artifactId>lz4-java</artifactId>
+      <version>1.8.0</version>
+    </dependency>
+    <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-bigquery</artifactId>
       <exclusions>
--- a/bigquery-connector-common/pom.xml
+++ b/bigquery-connector-common/pom.xml
@@ -31,6 +31,7 @@
       <groupId>org.lz4</groupId>
       <artifactId>lz4-java</artifactId>
       <version>1.8.0</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
--- a/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
+++ b/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
@@ -51,6 +51,7 @@ import com.google.cloud.bigquery.connector.common.ReadSessionCreatorConfig;
 import com.google.cloud.bigquery.connector.common.ReadSessionCreatorConfigBuilder;
 import com.google.cloud.bigquery.storage.v1.ArrowSerializationOptions.CompressionCodec;
 import com.google.cloud.bigquery.storage.v1.DataFormat;
+import com.google.cloud.bigquery.storage.v1.ReadSession.TableReadOptions.ResponseCompressionCodec;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
@@ -125,6 +126,10 @@ public class SparkBigQueryConfig
   static final CompressionCodec DEFAULT_ARROW_COMPRESSION_CODEC =
       CompressionCodec.COMPRESSION_UNSPECIFIED;
 
+  @VisibleForTesting
+  static final ResponseCompressionCodec DEFAULT_RESPONSE_COMPRESSION_CODEC =
+      ResponseCompressionCodec.RESPONSE_COMPRESSION_CODEC_UNSPECIFIED;
+
   static final String GCS_CONFIG_CREDENTIALS_FILE_PROPERTY =
       "google.cloud.auth.service.account.json.keyfile";
   static final String GCS_CONFIG_PROJECT_ID_PROPERTY = "fs.gs.project.id";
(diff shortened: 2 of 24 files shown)


bigquery-connector-common/pom.xml
```diff
@@ -27,6 +27,11 @@
       <groupId>com.google.api.grpc</groupId>
       <artifactId>proto-google-cloud-bigquerystorage-v1</artifactId>
     </dependency>
+    <dependency>
```
davidrabinowitz, line 30: lz4-java is provided by spark. If we can use that, then this dependencies can be removed. If we need version 1.8.0 then we need to shade it as well - the `net/jpountz/...` package as well as the `.so` library. Although, that the [native library loading](https://github.com/lz4/lz4-java/blob/master/src/java/net/jpountz/util/Native.java#L98) doesn't really support multiple versions, so perhaps using the spark provided version is the better option.

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
```diff
@@ -220,6 +226,7 @@ public static WriteMethod from(@Nullable String writeMethod) {
   private long readSessionCacheDurationMins = 5L;
   private SparkBigQueryProxyAndHttpConfig sparkBigQueryProxyAndHttpConfig;
   private CompressionCodec arrowCompressionCodec = DEFAULT_ARROW_COMPRESSION_CODEC;
+  private ResponseCompressionCodec responseCompressionCodec = DEFAULT_RESPONSE_COMPRESSION_CODEC;
```
davidrabinowitz, line 229: SparkBigQueryConfig is a Serializable class, and this enum is not Serializable. The field should be a `String`, converted to a ResponseCompressionCodec only on the executor.

davidrabinowitz (changes requested):
Please update the CHANGES.md file as well.

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
yqiu2: I'm not sure I understand why CompressionCodec is an serializable enum, but ResponseCompressionCodec is not a serializable enum.

bigquery-connector-common/pom.xml
yqiu2: When I remove org.lz4 lz4-java  from bigquery-connector-common/pom.xml, then I get this compilation in the bigquery-connector-common file

```
[ERROR] ~/spark-bigquery-connector/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/DecompressReadRowsResponse.java:[8,23] package net.jpountz.lz4 does not exist
```

I think this is because although [lz4-java](https://mvnrepository.com/artifact/org.lz4/lz4-java) is included in [spark-core](https://mvnrepository.com/artifact/org.apache.spark/spark-core), that isn't included in the bigquery-connector-common/pom.xml.  
I do see org.apache.spark spark-sql_2.13 and org.apache.spark spark-mllib_2.13 in spark-bigquery-connector-common/pom.xml, but I want to use lz4 in bigquery-connector-common and not spark-bigquery-connector-common code. 
plus, neither [spark-sql_2.13](https://mvnrepository.com/artifact/org.apache.spark/spark-sql_2.13/3.5.0) nor [spark-mllib_2.13](https://mvnrepository.com/artifact/org.apache.spark/spark-mllib) include lz4-java. 
So it seems like currently, bigquery-connector-common doesn't contain a spark sql or spark ml lib dependency containing a lz4-java dependency.

yqiu2: we decided to mark lz4-java as a provided dependency for bigquery-connector-common

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
davidrabinowitz: Looking at `ResponseCompressionCodec` inheritance tree I cannot see Serializable and therefore we need to convert it to another type (int/String). Alternatively, add a test to SparkBigQueryConfigTest verifying it it still Serializable

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/direct/BigQueryRDDContext.java
```diff
@@ -117,7 +117,8 @@ public scala.collection.Iterator<InternalRow> compute(Partition split, TaskConte
               readSession.getAvroSchema().getSchema(),
               Optional.of(schema),
               Optional.of(tracer),
-              SchemaConvertersConfiguration.from(options));
+              SchemaConvertersConfiguration.from(options),
+              options.getResponseCompressionCodec());
     } else {
       converter =
           ReadRowsResponseToInternalRowIteratorConverter.arrow(
```
davidrabinowitz, line 124: Shouldn't this be enabled to the arrow type as well?

spark-bigquery-dsv2/spark-bigquery-dsv2-common/src/main/java/com/google/cloud/spark/bigquery/v2/context/ArrowInputPartitionContext.java
```diff
@@ -48,6 +49,7 @@ public class ArrowInputPartitionContext implements InputPartitionContext<Columna
   private final ByteString serializedArrowSchema;
   private final com.google.common.base.Optional<StructType> userProvidedSchema;
   private final SparkBigQueryReadSessionMetrics sparkBigQueryReadSessionMetrics;
+  private final ResponseCompressionCodec responseCompressionCodec;
```
davidrabinowitz, line 52: This call is Serializable and therefore cannot have an un-Serializable member

davidrabinowitz (changes requested):

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
yqiu2: I already have  testDefaults()  in SparkBigQueryConfigTest.  is that insufficient?

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/direct/BigQueryRDDContext.java
yqiu2: I am decompressing in ReadRowsResponseInputStreamEnumeration.  Do I also need to do it here, or would that be attempting decompression twice?

yqiu2 (commented):
for arrow record batches, `response.getArrowRecordBatch().getSerializedRecordBatch()` is a ByteString.  
is converting  ByteString ==> byte[] ==>  InputStream less efficient than converting 
 ByteString ==> InputStream directly?

davidrabinowitz (approved):

davidrabinowitz (approved):

PR #1225, opened 2024-05-14
Fix multi-release jar shading

vishalkarve15:


vishalkarve15:
If we try to include the connector jar in maven-shade-plugin, it breaks the build with this error: `[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.2.3:shade (default) on project core: Error creating shaded jar: Problem shading JAR /Users/vkarve/.m2/repository/com/google/cloud/spark/spark-bigquery-with-dependencies_2.12/0.33.0/spark-bigquery-with-dependencies_2.12-0.33.0.jar entry META-INF/versions/17/com/fasterxml/jackson/core/io/doubleparser/FastDoubleSwar.class: java.lang.IllegalArgumentException: Unsupported class file major version 61 -> [Help 1]`  
Steps to reproduce:
- create a parent maven project with one child module
- in the child module, add dependency com.google.cloud.spark:spark-bigquery-with-dependencies_2.12 >= 0.33.0
- add maven-shade-plugin with at least 1 relocation rule (can be no-op)

vishalkarve15:
This happens because starting 0.33.0, we upgraded jackson-core from 2.14.x to 2.15.x is a adds Multi-release which adds `META-INF/versions/**`

davidrabinowitz (approved):

PR #1236, opened 2024-05-28
Adding shading verification

davidrabinowitz:
Adding a script to verify that the shaded artifacts contain no unshaded classes


spark-bigquery-parent/pom.xml
```diff
@@ -238,6 +239,18 @@
                 <version>${openlineage-spark.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
```
vishalkarve15, line 242: why do we need these dependencies?

davidrabinowitz: OpenTelemetry needs to be provided by the environment, as here we have just the API but not the reporting library. If we don't specify it as provided, it will be either (a) shaded, and therefor won't work with the reporter; or (b) unshaded, potentially colliding with the environment's OpenTelemetry. The best option is to avoid including it.

vishalkarve15: Got it. From offline discussion: needed by BQ storage library.

vishalkarve15 (approved):

Issue #1043, opened 2023-08-18
Indirect write drops column nullability and policy tags


khaledh:
We noticed that while this issue is resolved for direct writes (#1028), it still happens when we use indirect writes; the policy tags are dropped after the indirect write. We also noticed that non-nullable columns are created as nullable when using indirect write. We're trying to work around this issue, but we're hoping it would be fixed in the connector itself.

We tried both parquet and avro intermediate formats.

_(Original comment: https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/1028#issuecomment-1682710667)_

khaledh:
I updated the title/description to remove the issue about nullability. It turns out to be an issue on our side. The policy tags, however, are still not populated as expected with indirect writes.

isha97:
The fix will be available in the next release.

PR #1250, opened 2024-06-25
Issue #1043: fixed indirect write overwrite dropping policy tags

vishalkarve15:
Overwriting a table using writeMethod=indirect was causing policy tags to be dropped from the existing table.
This change fixes the issue.


isha97:
This is duplicate of https://github.com/GoogleCloudDataproc/spark-bigquery-connector/pull/1249

Issue #1254, opened 2024-07-09
NullPointerException when using Java connector with time partitioned tables


rafalh:
When I try to run the following code in PySpark in DataProc 2.2 I get NullPointerException from the connector:
```
dt = '2024-07-07'
df1 = (
    spark.read
    .format("bigquery")
    .option("table", 'table1')
    .option("filter", f"_PARTITIONDATE = '{dt}'")
    .load()
    .select("page_view.page_view_id")
)
df2 = (
    spark.read
    .format("bigquery")
    .option("table", 'table2')
    .option("filter", f"_PARTITIONDATE = '{dt}'")
    .load()
    .select("page_view_id")
)
df1.join(df2, "page_view_id").cache()
```

I tried different versions of the connector (setting `SPARK_BQ_CONNECTOR_URL` metadata for DataProc cluster):
* spark-bigquery-with-dependencies_2.12-0.34.0.jar - works (0.34.0 is the default version for DataProc 2.2)
* spark-3.2-bigquery-0.34.0.jar - fails
* spark-3.1-bigquery-0.34.0.jar - works
* spark-3.5-bigquery-0.39.1.jar - fails

It seems only Java connector for Spark 3.2+ is affected. It makes sense because `Spark32BigQueryScanBuilder` can be seen in the stacktrace. From what I understand the exception is caused by `StandardTableDefinition.getTimePartitioning()` returning `null` which makes sense because my tables has no explicit partitioning column (`_PARTITIONTIME` pseudo-column is used).

Stacktrace:
```
Py4JJavaError: An error occurred while calling o104.cache.
: org.apache.spark.SparkException: [INTERNAL_ERROR] The Spark SQL phase optimization failed with an internal error. You hit a bug in Spark or the Spark plugins you use. Please, report this bug to the corresponding communities or vendors, and provide the full stack trace.
	at org.apache.spark.SparkException$.internalError(SparkException.scala:107)
	at org.apache.spark.sql.execution.QueryExecution$.toInternalError(QueryExecution.scala:537)
	at org.apache.spark.sql.execution.QueryExecution$.withInternalError(QueryExecution.scala:549)
	at org.apache.spark.sql.execution.QueryExecution.$anonfun$executePhase$1(QueryExecution.scala:219)
	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:900)
	at org.apache.spark.sql.execution.QueryExecution.executePhase(QueryExecution.scala:218)
[...]

PR #1258, opened 2024-07-12
Issue #1254: fix getting partitioning fields for pseudo columns

vishalkarve15:
When getting partitioning fields, `timePartitioning.getField()` returns null when the table is ingestion time partitioned (partitioned by `_PARTITIONTIME` OR `_PARTITIONDATE`).
The change fixes this bug.

### DIFF
--- a/spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/ReadIntegrationTestBase.java
+++ b/spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/ReadIntegrationTestBase.java
@@ -40,6 +40,7 @@ import java.util.stream.Stream;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Encoders;
 import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.expressions.SparkVersion;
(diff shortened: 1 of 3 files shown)


davidrabinowitz (changes requested):
Can you please add an integration test based on issue #1254 ?

vishalkarve15:
Added integration tests which fail as expected without the fix.

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/ReadIntegrationTestBase.java
```diff
@@ -713,4 +714,42 @@ public void testPushDateTimePredicate() {
             .where("orderDateTime < '2023-10-25 10:00:00'");
     assertThat(df.count()).isEqualTo(2);
   }
+
+  @Test
```
davidrabinowitz, line 718: Please add some data into the tables, and we need to heave at least one assert. Also, as this refers to ingestion time partitioning, then the partition date should be taken from the test date (dynamically), or to push data into a specific partition using table decorators.

davidrabinowitz (changes requested):

isha97 (approved):

PR #1259, opened 2024-07-12
Encode snapshotTimeMillis in view materialization query

tom-s-powell:
`snapshotTimeMillis` does not currently make sense with views since the view is materialized to a table without taking this property into consideration, and then a read session is created attempting to read the materialized table at this timestamp (which has only just been created). This change will instead use `snapshotTimeMillis` to add `FOR SYSTEM_TIME AS OF TIMESTAMP_MILLIS` clause to the query that materializes the view to a table and the read session simply consumes that.


bigquery-connector-common/src/test/java/com/google/cloud/bigquery/connector/common/ReadSessionCreatorTest.java
```diff
@@ -27,12 +27,7 @@
 import com.google.api.gax.grpc.testing.MockGrpcService;
 import com.google.api.gax.grpc.testing.MockServiceHelper;
 import com.google.api.gax.rpc.UnaryCallable;
-import com.google.cloud.bigquery.Field;
-import com.google.cloud.bigquery.Schema;
-import com.google.cloud.bigquery.StandardSQLTypeName;
-import com.google.cloud.bigquery.StandardTableDefinition;
-import com.google.cloud.bigquery.TableId;
-import com.google.cloud.bigquery.TableInfo;
+import com.google.cloud.bigquery.*;
```
isha97, line 30: nit : please expand all imports

isha97 (changes requested):

bigquery-connector-common/src/test/java/com/google/cloud/bigquery/connector/common/ReadSessionCreatorTest.java
tom-s-powell: Fixed

isha97 (approved):

PR #1281, opened 2024-08-23
Configure alternative BigNumeric precision and scale defaults

isha97:


README.md
```diff
@@ -908,6 +908,26 @@ word-break:break-word
      </td>
      <td>Read</td>
```
davidrabinowitz, line 909: `README.md` should not be edited, just `README-template.md`

README.md
```diff
@@ -908,6 +908,26 @@ word-break:break-word
...
    </tr>
+  <tr>
+     <td><code>bigNumericFieldsPrecision</code>
+     </td>
+     <td>A config to specify the precision to use while reading all the BigNumeric fields of a table.
```
davidrabinowitz, line 914: Should be `An alternative default precision for BigNumeric fields, as the BigQuery default is too wide for Spark. Values can be between 1 and 38.  This default is used only when the field has an unparameterized BigNumeric type.`

README.md
```diff
@@ -908,6 +908,26 @@ word-break:break-word
...
+          Please note that there might be data loss if the actual data's precision is more than what is specified.
+          By default this is not set and precision from the field is used.
+         <br/> (Optional)
+     </td>
+     <td>Read</td>
+   </tr>
+  <tr>
+     <td><code>bigNumericFieldsScale</code>
+     </td>
+     <td>A config to specify the scale to use while reading all the BigNumeric fields of a table.
```
davidrabinowitz, line 924: Should be `An alternative default scale for BigNumeric fields. Values can be between 0 and 38, and less then bigNumericFieldsPrecision.  This default is used only when the field has an unparameterized BigNumeric type.`

davidrabinowitz (changes requested):
Please update the CHANGES.md file

README.md
isha97: Fixed

README.md
isha97: Fixed

README.md
isha97: Fixed

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
```diff
@@ -167,6 +168,9 @@ public static WriteMethod from(@Nullable String writeMethod) {
 
   public static final String GPN_ATTRIBUTION = "GPN";
 
+  public static final String BIG_NUMERIC_DEFAULT_PRECISION = "bigNumericDefaultPrecision";
```
davidrabinowitz, line 171: The parameter names should match the documentation (and exceptions)

davidrabinowitz (approved):

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
isha97: Done

Issue #1290, opened 2024-09-05
Issue with BigQuery counts in Spark after writing to table


rchynoweth:
I am having an issue getting accurate counts when reading/writing to BigQuery from Databricks after installing the connector. 

Connector Version: spark-3.5-bigquery-0.39.1.jar
Apache Spark 3.5.0
Scala 2.12 
Databricks 14.3LTS

Code to replicate: 
```
from pyspark.sql import Row

# Create a list of Rows
data = [
    Row(col1="value1_1", col2="value1_2", col3="value1_3"),
    Row(col1="value2_1", col2="value2_2", col3="value2_3"),
    Row(col1="value3_1", col2="value3_2", col3="value3_3"),
    Row(col1="value4_1", col2="value4_2", col3="value4_3"),
    Row(col1="value5_1", col2="value5_2", col3="value5_3"),
    Row(col1="value6_1", col2="value6_2", col3="value6_3"),
    Row(col1="value7_1", col2="value7_2", col3="value7_3"),
    Row(col1="value8_1", col2="value8_2", col3="value8_3"),
    Row(col1="value9_1", col2="value9_2", col3="value9_3"),
    Row(col1="value10_1", col2="value10_2", col3="value10_3")
]

# Create DataFrame
df = spark.createDataFrame(data)

# count table
(spark.read
  .format("bigquery")
  .option("table", 'mydataset.test_read_write_table')
  .option("project", 'myproject')
  .load()
  ).count()

### prints 50

# write 10 rows to table
(df.write
  .mode('append')
  .format("bigquery")
  .option("project", 'myproject')
  .option("writeMethod", "direct")
  .save('mydataset.test_read_write_table')
)

# check count again
(spark.read
  .format("bigquery")
  .option("table", 'mydataset.test_read_write_table')
  .option("project", 'myproject')
  .load()
).count()
### prints 50 but should be 60
```

When I check the BQ table the rows are updated but not reflected in my Dataframe. If I use the query option instead of the table and perform "select count(1) from mydataset.test_read_write_table" then the counts are accurate. This seems like a potential cache problem which I tried using the cacheExpirationTimeInMinutes option to 0 but it seems to not work. However, if I set it to a positive integer it does work after the time setting is up.

agrawal-siddharth:
What happens if you omit the first count check? I'm trying to narrow down if there is an issue with the spark.read being called twice.

rchynoweth:
[...]

PR #1293, opened 2024-09-16
Issue #1290: Stopped using metadata for optimized count path

vishalkarve15:
TableInfo.getNumRows() only covers the data that's materialized to data files, without including the data in streaming buffer. The data in streaming buffer (a.k.a, ingested through Write API, but not converted to data file yet) is readable using ReadAPI, but accessing TableInfo.getNumRows() won't reflect whatever changes are made in the streaming buffer. Running a select count(1) query works, as it triggers actual data scan through ReadAPI.

This issue only affects direct write.

### DIFF
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
@@ -15,8 +15,10 @@
  */
 package com.google.cloud.bigquery.connector.common;
 
+import static com.google.cloud.bigquery.connector.common.BigQueryUtil.getPartitionFields;
 import static com.google.cloud.bigquery.connector.common.BigQueryUtil.getQueryForRangePartitionedTable;
 import static com.google.cloud.bigquery.connector.common.BigQueryUtil.getQueryForTimePartitionedTable;
+import static com.google.cloud.bigquery.connector.common.BigQueryUtil.isBigQueryNativeTable;
 
 import com.google.cloud.BaseServiceException;
 import com.google.cloud.RetryOption;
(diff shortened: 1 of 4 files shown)


spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
```diff
@@ -2664,6 +2664,34 @@ public void testTableDescriptionRemainsUnchanged() {
         .isEqualTo(bq.getTable(testDataset.toString(), testTable).getDescription());
   }
 
+  @Test
+  public void testDirectWriteCountAfterWrite() throws InterruptedException {
```
davidrabinowitz, line 2668: Why not to have a single `testCountAfterWrite()` ? do we need different logic for direct vs indirect writes?

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -604,9 +604,8 @@ public long calculateTableSize(TableId tableId, Optional<String> filter) {
 
   public long calculateTableSize(TableInfo tableInfo, Optional<String> filter) {
     TableDefinition.Type type = tableInfo.getDefinition().getType();
-    if (type == TableDefinition.Type.TABLE && !filter.isPresent()) {
```
davidrabinowitz, line 607: Is there a way to refresh `tableInfo`? `getNumRows()` is significantly faster

davidrabinowitz (changes requested):

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/WriteIntegrationTestBase.java
vishalkarve15: Right, missed that.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
vishalkarve15: Let me check.

vishalkarve15: As discussed offline, let's go with this approach for now. Can be revisited later if optimization is required.

davidrabinowitz (changes requested):
The integration test fails on tables which require partition filter

vishalkarve15:
We have 2 options here:
- fallback to TableInfo for requirePartitionFilter tables. But this preserves existing behavior with potential for incorrect count.
- use count(*) for requirePartitionFilter tables which fails if partition filter is not specified. This is a breaking change.

I'd prefer the 2nd option which makes it a breaking change, because it is the intended behavior of require partition filter.

vishalkarve15:
@davidrabinowitz wouldn't option 2 make more sense? Consider the dataframe that's querying a table with requirePartitionFilter = true.
```
Dataset<Row> readDF = spark.read().format("bigquery").load(table);
readDF.count(); // succeeds but ends up running count(*) on the entire table
readDF.collect(); // fails due to require partition filter
```

I think the purpose of require partition filter is for users to be able to avoid running expensive queries against the whole table. This seems to be enabling that behavior since we're changing it at the query level.

Existing behavior for Biglake tables is to error out. By having BQ native tables error out too, we can make this behavior consistent.

vishalkarve15:
As discussed offline, will go ahead with option 3 since we can't introduce a breaking change.

davidrabinowitz (approved):

PR #1340, opened 2025-02-23
fix: respect project id for incoming identifier

thomaschow:
- Need to respect the incoming project id as well.

### DIFF
--- a/spark-bigquery-dsv2/spark-3.5-bigquery-lib/src/main/java/com/google/cloud/spark/bigquery/BigQueryCatalog.java
+++ b/spark-bigquery-dsv2/spark-3.5-bigquery-lib/src/main/java/com/google/cloud/spark/bigquery/BigQueryCatalog.java
@@ -121,6 +121,7 @@ public class BigQueryCatalog implements TableCatalog {
   @Override
   public Table loadTable(Identifier identifier) throws NoSuchTableException {
     logger.debug("loading table [{}])", format(identifier));
+    TableId tableId = toTableId(identifier);
     try {
       return identifierToTableCache.get(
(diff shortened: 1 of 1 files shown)


spark-bigquery-dsv2/spark-3.5-bigquery-lib/src/main/java/com/google/cloud/spark/bigquery/BigQueryCatalog.java
```diff
@@ -130,7 +131,9 @@ public Table loadTable(Identifier identifier) throws NoSuchTableException {
                   Spark35BigQueryTable::new,
                   null,
                   ImmutableMap.of(
-                      "dataset", identifier.namespace()[0], "table", identifier.name())));
+                          "project", tableId.getProject(),
```
davidrabinowitz, line 134: I'm afraid it's not that trivial - project may by `null` in the TableId. Also, we need to add tests for cross project reading.

tchow-zlai: Is there an existing test suite for this class? I understand it's fairly new so no worries if not. I'm happy to add some if this change looks directionally sound.

thomaschow: Is there an existing test suite for this class? I understand it's fairly new so no worries if not. I'm happy to add some if this change looks directionally sound. Please advise!

PR #1393, opened 2025-07-15
Fix: Resolve nightly build and deploy failures

BoqianShi:
This PR fixes the nightly build by ensuring the Scala support module is correctly included in the build artifact. And, it also temporarily removes the deploy stage to avoid failure.

### DIFF
--- a/cloudbuild/nightly.sh
+++ b/cloudbuild/nightly.sh
@@ -97,6 +97,11 @@ case $STEP in
     exit
     ;;
 
+  deploy)
+#    $MVN deploy:deploy -DskipTests -Dscala.skipTests=true -Prelease-nightly,dsv1_2.12,dsv1_2.13,dsv2_3.1,dsv2_3.2,dsv2_3.3,dsv2_3.4,dsv2_3.5,dsv2_4.0
+    exit
+    ;;
+
   *)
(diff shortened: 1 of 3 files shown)


cloudbuild/nightly.yaml
```diff
@@ -22,11 +22,6 @@ steps:
     entrypoint: 'bash'
     args: ['/workspace/cloudbuild/nightly.sh', 'copy-to-gcs']
 
-  - name: 'gcr.io/$PROJECT_ID/dataproc-spark-bigquery-connector-nightly'
```
davidrabinowitz: Please keep this step

cloudbuild/nightly.sh
```diff
@@ -97,13 +97,6 @@ case $STEP in
     exit
     ;;
 
-  deploy)
```
davidrabinowitz, line 100: Please keep this step, comment out the actual task and add a TODO comment

BoqianShi: Done

cloudbuild/nightly.yaml
BoqianShi: Done

davidrabinowitz (approved):

PR #1396, opened 2025-07-18
Using the new central-publishing-maven-plugin

davidrabinowitz:
Based on the documentation in https://central.sonatype.org/publish/publish-portal-maven/


BoqianShi (approved):
LGTM

PR #1419, opened 2025-09-16
Fix the logging for pushdown filters

kaushikrv42:
Fix the call to get friendly name. Since BQ does not create a friendly name for a table when we use a builder to create a table, we need to use the big query util to get the friendly table name.

### DIFF
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/ReadSessionCreator.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/ReadSessionCreator.java
@@ -16,7 +16,7 @@
 package com.google.cloud.bigquery.connector.common;
 
 import static com.google.cloud.bigquery.connector.common.BigQueryErrorCode.UNSUPPORTED;
-import static com.google.cloud.bigquery.connector.common.BigQueryUtil.friendlyTableName;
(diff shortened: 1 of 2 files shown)


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/ReadSessionCreator.java
```diff
@@ -16,6 +16,7 @@
 package com.google.cloud.bigquery.connector.common;
 
 import static com.google.cloud.bigquery.connector.common.BigQueryErrorCode.UNSUPPORTED;
+import static com.google.cloud.bigquery.connector.common.BigQueryUtil.friendlyTableName;
```
davidrabinowitz: Please avoid static import of method name

kaushikrv42: Updated to use `BigQueryUtil.friendlyTableName` instead. I referred to other users of BigQueryUtil methods - there's a lot of static imports in the code.

davidrabinowitz (approved):

davidrabinowitz (changes requested):
Please add the read session filters to the BatchScan toString

davidrabinowitz (approved):

PR #1434, opened 2025-10-23
Support per-catalog configuration for project and location for BigQueryCatalog

BoqianShi:
This PR enhances the BigQueryCatalog by adding support for specifying a project and location directly in the Spark configuration. This allows users to easily register and interact with BigQuery datasets from different projects and locations without having to override the global parentProjectId.

This makes it easier to work with datasets in different projects and regions. For example, you can now read from a public dataset in one project and write to a table in a specific location (e.g., "US") in your own project.

How to Use
Users can now configure a BigQuery catalog with a specific project and location like this:
```// In your Spark session
spark.conf.set("spark.sql.catalog.my_catalog", "com.google.cloud.spark.bigquery.BigQueryCatalog")
spark.conf.set("spark.sql.catalog.my_catalog.projectId", "bigquery-public-data")
spark.conf.set("spark.sql.catalog.my_catalog.location", "US")
```

### DIFF
--- a/README-template.md
+++ b/README-template.md
@@ -433,10 +433,26 @@ word-break:break-word
    <td>Read/Write</td>
   </tr>
   <tr valign="top">
+   <td><code>billingProject</code>
+   </td>
+   <td>The Google Cloud Project ID to use for <strong>billing</strong> (API calls, query execution).
+       <br/>(Optional. Defaults to the project of the Service Account being used)
+   </td>
+   <td>Read/Write</td>
+  </tr>
+  <tr valign="top">
    <td><code>parentProject</code>
    </td>
-   <td>The Google Cloud Project ID of the table to bill for the export.
+   <td><strong>(Deprecated)</strong> Alias for <code>billingProject</code>.
        <br/>(Optional. Defaults to the project of the Service Account being used)
+   </td>
+   <td>Read/Write</td>
+  </tr>
+  <tr valign="top">
+   <td><code>location</code>
+   </td>
+   <td>The BigQuery location where the data resides (e.g. US, EU, asia-northeast1).
+       <br/>(Optional. Defaults to BigQuery default)
    </td>
    <td>Read/Write</td>
   </tr>


spark-bigquery-dsv2/spark-3.5-bigquery-lib/src/main/java/com/google/cloud/spark/bigquery/BigQueryCatalog.java
```diff
@@ -82,7 +82,11 @@ public class BigQueryCatalog implements TableCatalog, SupportsNamespaces {
   @Override
   public void initialize(String name, CaseInsensitiveStringMap caseInsensitiveStringMap) {
     logger.info("Initializing BigQuery table catalog [{}])", name);
-    Injector injector = new InjectorBuilder().withTableIsMandatory(false).build();
+    Injector injector =
+        new InjectorBuilder()
+            .withOptions(caseInsensitiveStringMap.asCaseSensitiveMap())
+            .withTableIsMandatory(false)
+            .build();
```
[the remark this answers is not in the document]
BoqianShi: Not true. BigQuery table's location are inherited from the dataset, and the location is passed to the dataset via BigQueryOptions if available.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -975,6 +975,7 @@ public boolean datasetExists(DatasetId datasetId) {
 
   public void createDataset(DatasetId datasetId, Map<String, String> metadata) {
     DatasetInfo.Builder datasetInfo = DatasetInfo.newBuilder(datasetId);
+    Optional.ofNullable(bigQuery.getOptions().getLocation()).ifPresent(datasetInfo::setLocation);
```
davidrabinowitz, line 981: Please add  comment for this line explaining why it is there

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
```diff
@@ -424,6 +426,8 @@ public static SparkBigQueryConfig from(
 
     config.parentProjectId =
         getAnyOption(globalOptions, options, "parentProject").or(defaultBilledProject());
+    config.catalogProjectId = getOption(options, "projectId");
+    config.catalogLocation = getOption(options, "location");
```
davidrabinowitz: Is the parameter `location` or `bigquery_location`?

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
```diff
@@ -206,6 +206,8 @@ public static WriteMethod from(@Nullable String writeMethod) {
   com.google.common.base.Optional<String> temporaryGcsBucket = empty();
   com.google.common.base.Optional<String> persistentGcsBucket = empty();
   com.google.common.base.Optional<String> persistentGcsPath = empty();
+  com.google.common.base.Optional<String> catalogProjectId = empty();
```
davidrabinowitz, line 212: add comment why do we need a separate set of properties

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientModule.java
```diff
@@ -106,11 +106,16 @@ public BigQueryClient provideBigQueryClient(
     BigQueryOptions.Builder options =
         BigQueryOptions.newBuilder()
             .setHeaderProvider(headerProvider)
-            .setProjectId(config.getParentProjectId())
             .setCredentials(bigQueryCredentialsSupplier.getCredentials())
             .setRetrySettings(config.getBigQueryClientRetrySettings())
             .setUniverseDomain(bigQueryCredentialsSupplier.getUniverseDomain());
 
+    // Determine the project for data location (the catalog's project, if specified)
```
davidrabinowitz: What happens if this is used in non-catalog scenario?

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
BoqianShi: In a non-catalog scenario, both BigQueryOptions.quotaProjectId and BigQueryOptions.projectId will be set to the parent projectId, as before.

BoqianShi: Added

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
BoqianShi: Added

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/SparkBigQueryConfig.java
BoqianShi: Changed

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClientModule.java
BoqianShi: In a non-catalog scenario (e.g., standard `spark.read.format("bigquery")...`), the configuration typically uses `project` (for the table's project) and `parentProject` (for the billing project), but not `projectId`.

Therefore, `config.getCatalogProjectId()` will return `Optional.empty()`, and the logic successfully falls back to `.orElse(config.getParentProjectId())`. This maintains existing behavior for standard jobs, ensuring `parentProjectId` (derived from `parentProject` or the default credential project) is used for billing.

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/CatalogIntegrationTestBase.java
```diff
@@ -40,8 +43,31 @@ public class CatalogIntegrationTestBase {
...
+  public static void setupSparkSession() {
+    spark =
+        SparkSession.builder()
+            .appName("catalog test")
+            .master("local[*]")
+            .config("spark.sql.legacy.createHiveTableByDefault", "false")
+            .config("spark.sql.sources.default", "bigquery")
+            .config("spark.datasource.bigquery.writeMethod", "direct")
+            .config("spark.sql.defaultCatalog", "bigquery")
+            .config("spark.sql.catalog.bigquery", "com.google.cloud.spark.bigquery.BigQueryCatalog")
```
yalimu-g, line 60: any other configs for spark.sql.catalog.bigquery besides "com.google.cloud.spark.bigquery.BigQueryCatalog"?

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/CatalogIntegrationTestBase.java
```diff
@@ -238,10 +244,136 @@ public void testDropDatabase() {
...
+  @Test
+  public void testCatalogInitializationWithProject() {
+    try {
+      spark
+          .conf()
+          .set(
+              "spark.sql.catalog.public_catalog",
+              "com.google.cloud.spark.bigquery.BigQueryCatalog");
+      // Use 'projectId' instead of 'project' - this is the correct property name
+      spark.conf().set("spark.sql.catalog.public_catalog.projectId", "bigquery-public-data");
```
yalimu-g, line 277: is different parent project and catalog project a valid use case?

spark-bigquery-dsv2/spark-3.5-bigquery-lib/src/main/java/com/google/cloud/spark/bigquery/BigQueryCatalog.java
```diff
@@ -152,6 +168,11 @@ public Table loadTable(Identifier identifier) throws NoSuchTableException {
 
   Map<String, String> toLoadProperties(Identifier identifier) {
     ImmutableMap.Builder<String, String> result = ImmutableMap.builder();
+    Map<String, String> processedOptions = new java.util.HashMap<>(options);
```
yalimu-g, line 171: would this be confusing to swap parameters' name?

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/CatalogIntegrationTestBase.java
BoqianShi: For a basic setup using the default credentials and project, setting the catalog implementation class is sufficient.

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/CatalogIntegrationTestBase.java
BoqianShi: Yes, this is a part of the reason why we need this PR. These two are different in terms of:
 
- **Catalog Project (Data)**: Where the tables reside (e.g., `bigquery-public-data`).
- **Parent Project (Billing)**: Who pays for the query execution (e.g., the user's project).

Without this separation, we cannot query public datasets because we cannot bill operations to the `bigquery-public-data` project (we don't have permissions). We must read from `bigquery-public-data` but bill to `my-auth-project`.

spark-bigquery-dsv2/spark-3.5-bigquery-lib/src/main/java/com/google/cloud/spark/bigquery/BigQueryCatalog.java
BoqianShi: It might seem counter-intuitive at first, but it corrects a semantic mismatch between Spark's Catalog configuration and the BigQuery Connector's internal options.

- **`projectId`**: In the connector's lower-level config, this key often forces the **Billing Project** context.
- **`project`**: This key is used to define the **Table/Data Location**.

When a user sets `spark.sql.catalog.mycat.projectId`, they intend to define where the *data* is ("this catalog is for project X"). If we pass this as `projectId` to the connector, it mistakenly switches the billing context too. Renaming it to `project` ensures we respect the user's data location intent without breaking the billing configuration (which should fallback to the authenticated user's project).

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/CatalogIntegrationTestBase.java
```diff
@@ -40,8 +43,31 @@ public class CatalogIntegrationTestBase {
 
   BigQuery bigquery = IntegrationTestUtils.getBigquery();
 
+  protected static SparkSession spark;
   private String testTable;
 
+  @BeforeClass
```
davidrabinowitz, line 50: Doing that means that all tests will share the same Spark Session and same catalog. This is why each test create its own SparkSession

BoqianShi: SG. I changed the test to use globalSpark.newSession(), which ensures test isolation while avoiding the instability caused by frequently restarting the SparkContext.

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/CatalogIntegrationTestBase.java
yalimu-g: I see. Thanks for the information.

spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/CatalogIntegrationTestBase.java
```diff
@@ -40,8 +43,47 @@ public class CatalogIntegrationTestBase {
 
   BigQuery bigquery = IntegrationTestUtils.getBigquery();
 
+  protected SparkSession spark;
+  private static SparkSession globalSpark;
   private String testTable;
 
+  @BeforeClass
+  public static void setupGlobalSpark() {
```
davidrabinowitz, line 51: Why globalSpark is needed in the first place? Why not to create a new session on the `@Before` and `close()` it on the `@After`?

BoqianShi: We initially tried creating and closing the SparkSession in `@Before`/`@After` for strict isolation. However, calling `spark.stop()` shuts down the underlying `SparkContext`. Restarting the `SparkContext` repeatedly within the same JVM caused flaky `IllegalStateException: Cannot call methods on a stopped SparkContext` errors during testing.

By using `globalSpark` (to hold the `SparkContext`) and calling `globalSpark.newSession()` in `@Before`, we achieve the best of both worlds:
1.  **Isolation**: Each test gets a fresh [SparkSession](cci:1://file:///Users/boqianshi/workspace/dataproc/spark-connector/private_connector/catalog-project-id/spark-bigquery-connector/spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/CatalogIntegrationTestBase.java:71:2-79:3) with its own `SessionState`, `SQLConf`, and catalog configuration.
2.  **Stability**: We avoid the overhead and instability of repeatedly restarting the `SparkContext`.

BoqianShi: Readme added with property descriptions.

Issue #1456, opened 2026-01-10
docs: Document schema preservation behavior on SaveMode.Overwrite (breaking change from 0.22.2 to 0.41.0)


sahilkumarsingh:
## Summary

The behavior of `SaveMode.Overwrite` underwent significant changes between versions 0.22.2 and 0.41.0, yet this modification is not documented. Users upgrading between these versions may face unexpected failures when overwriting tables with differing column types.

## Problem

In Spark-BigQuery connector v0.22.2, using mode("overwrite") allows updating BigQuery table columns if the data type of the dataframe being overwritten differs from that of the existing table. However, in v0.41.0, this operation fails because of stricter schema update rules, resulting in errors such as:

Samplecode:
```
df.write.format("bigquery") \
  .mode("overwrite") \
  .option("temporaryGcsBucket", bucketName) \
  .option("table", table) \
  .option("project", project_id) \
  .option("parentProject", parentProject_id) \
  .save()
```
Error:
`BigQueryException: Error while reading data, error message: Parquet column 'id' has type DOUBLE which does not match the target cpp_type INT64.`

I noticed this behavioural change in Databricks during the upgrade from DBR 13.3 to DBR 16.4.

DBR 13 supports connector version **0.22.2**.
DBR 16.1 supports connector version **0.41.0**.

Ref: Support for connector version **0.41.0** began with DBR 16.1, whereas earlier versions supported **0.22.2**. More details can be found at https://docs.databricks.com/aws/en/archive/runtime-release-notes/16.1#bigquery-connector-upgrade

## Impact

- Users upgrading from older versions encounter unexpected failures.
- CHANGES.md lacks a warning about this behavioral change.
- README.md does not explain the overwrite schema behavior.
- No migration guide for major version upgrades.

## Can we include this information in the README.md or another documentation source?

sahilkumarsingh:
Hi @davidrabinowitz  @yalimu-g , does this change add value?

Do let me know if this is not required. thanks!

yalimu-g:
Hi @sahilkumarsingh have you tried the config `allowFieldRelaxation`?

PR #1457, opened 2026-01-10
docs: Document schema preservation behavior on SaveMode.Overwrite (#1456)

sahilkumarsingh:
## Description

This PR addresses the issue [#1456](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/issues/1456)

Documents the schema preservation behaviour when using `SaveMode.Overwrite`, which was updated somewhere between 0.22.2 and 0.41.0 versions.

## Changes

- Added "Schema Behaviour on Overwrite" section to README-template.md
- Explains that the existing table schema is preserved during overwrites
- Documents workarounds for intentional schema changes

## Why

Users upgrading from versions prior to 0.41.0 encounter unexpected failures when overwriting tables with different column types. This documentation helps users understand the expected behaviour and provides workarounds.

### DIFF
--- a/README-template.md
+++ b/README-template.md
@@ -375,7 +375,7 @@ df.writeStream \
 
 #### Schema Behavior on Overwrite
 
-When using `SaveMode.Overwrite` (`.mode("overwrite")`), the connector **preserves the existing table's schema**. 
+When using `SaveMode.Overwrite` (`.mode("overwrite")`), the connector **preserves the existing table's schema**.
 The data is truncated, but column types, descriptions, and policy tags are retained.
 
 ```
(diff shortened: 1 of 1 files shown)


sahilkumarsingh:
cc: @yalimu-g

README-template.md
```diff
@@ -373,6 +373,28 @@ df.writeStream \
...
+  .mode("overwrite") \
+  .option("temporaryGcsBucket","some-bucket") \
+  .save("dataset.table")
+```
+
+**Important:** If your DataFrame has a different schema than the existing table (e.g., changing a column from 
+`INTEGER` to `DOUBLE`), the write will fail with a type mismatch error. To change the schema, either:
+- Drop the table before overwriting
+- Use BigQuery DDL to alter the table schema first
+
```
yalimu-g, line 393: For some of the schema difference the following options can work with overwrite:
Programmatic Relaxation: Set .option("allowFieldRelaxation", "true") for nullability changes and .option("allowFieldAddition", "true") for new columns.

sahilkumarsingh: Yes, we can add this detail. Shall I add it?

yalimu-g: yes please

sahilkumarsingh: Done, please check.

Issue #1458, opened 2026-01-14
Class Cast Exception with Spark 4, using preview driver (0.43.1-preview)


Nassiel:
Hi,

While doing a super simple query using Spark 4.0.1 and BigQuery with the 43.1-preview for spark 4.0:
```
# BigQuery SQL query
query = """
SELECT mac, ACCOUNT.num
FROM `project.collection.table` 
WHERE data_date = "2026-01-11"
"""

df = (spark.read.format("bigquery")
    .option("viewsEnabled", "true")
    .option("materializationProject", "project")
    .load(query)
     )

df.show()
```

Is throwing a cast error internally:

```
Caused by: java.lang.ClassCastException: cannot assign instance of com.google.cloud.spark.bigquery.repackaged.com.google.common.collect.RegularImmutableMap to field com.google.cloud.bigquery.connector.common.QueryParameterHelper.namedParameters of type com.google.common.collect.ImmutableMap in instance of com.google.cloud.bigquery.connector.common.QueryParameterHelper
```

I tried to find workarounds, but nothing has worked.

Nassiel:
> Hi [@Nassiel](https://github.com/Nassiel) , can you please provide the full stack trace of the Exception?

Here is the Spark Connect exception @davidrabinowitz 

```
26/01/16 18:36:09 INFO TaskSetManager: Starting task 0.0 in stage 4179.0 (TID 93580) (10.170.32.69,executor 677, partition 0, PROCESS_LOCAL, 29603 bytes)                                                                                        │
│     at org.apache.spark.sql.execution.QueryExecution$.withInternalError(QueryExecution.scala:654)                                                                                                                                                
│     at org.apache.spark.sql.classic.Dataset.$anonfun$withAction$1(Dataset.scala:2232)                                                                                                                                                            
│     at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId0$8(SQLExecution.scala:163)                                                                                                                                       
│     at org.apache.spark.sql.execution.SQLExecution$.withSessionTagsApplied(SQLExecution.scala:272)                                                                                                                                               
│     at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId0$7(SQLExecution.scala:125)                                                                                                                                       
│     at org.apache.spark.JobArtifactSet$.withActiveJobArtifactState(JobArtifactSet.scala:94)                                                                                                                                                      
[...]

PR #1459, opened 2026-01-16
Use Java Standard Interface to fix the cast exception

yalimu-g:
The com.google.cloud.bigquery.connector.common package should be relocated along with the rest of the dependencies. This ensures that the code within the shaded connector strictly uses its own internal, shaded versions of the classes, avoiding conflicts with any unshaded libraries on the classpath.

If this fixes the issue of #1458 , I will remove the commented code.

### DIFF
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/QueryParameterHelper.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/QueryParameterHelper.java
@@ -18,9 +18,10 @@ package com.google.cloud.bigquery.connector.common;
 import com.google.cloud.bigquery.QueryJobConfiguration;
 import com.google.cloud.bigquery.QueryParameterValue;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


spark-bigquery-parent/pom.xml
```diff
@@ -628,7 +628,8 @@
                             <pattern>com.google</pattern>
                             <shadedPattern>com.google.cloud.spark.bigquery.repackaged.com.google</shadedPattern>
                             <excludes>
-                                <exclude>com.google.cloud.bigquery.connector.common.**</exclude>
+                                <!-- the bigquery connector common library needs to be repackaged -->
+                                <!-- <exclude>com.google.cloud.bigquery.connector.common.**</exclude> -->
```
davidrabinowitz: I'm ont sure this is the right fix, also there can be other libraries (such as OpenLineage) that may depend on this package

yalimu-g: Revert this change.

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/QueryParameterHelper.java
```diff
@@ -36,17 +36,17 @@ public final class QueryParameterHelper implements Serializable {
...
       ParameterMode mode,
-      ImmutableMap<String, QueryParameterValue> named,
-      ImmutableList<QueryParameterValue> positional) {
+      Map<String, QueryParameterValue> named,
+      List<QueryParameterValue> positional) {
     this.mode = mode;
 
-    this.namedParameters = named;
-    this.positionalParameters = positional;
+    this.namedParameters = ImmutableMap.copyOf(named);
```
davidrabinowitz: The Guava class may still have a serialization issue, as it is going to be the class in runtime. We can replace it with `Collections.unmodifiableMap()`

PR #1460, opened 2026-01-20
Remove SparkSession from class to prevent it being included in the serialization process

yalimu-g:


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/metrics/SparkBigQueryReadSessionMetrics.java
```diff
@@ -41,7 +41,7 @@ public class SparkBigQueryReadSessionMetrics extends SparkListener
   private final LongAccumulator scanTimeAccumulator;
   private final LongAccumulator parseTimeAccumulator;
   private final String sessionId;
-  private final SparkSession sparkSession;
+  private final transient SparkSession sparkSession;
```
davidrabinowitz: Given the usage of the SparkSession, we need to either: 

1. Implement a readObject() method that will initialize it to SaprkSession.active()
2. Given that the only usage is in onJobEnd and in the constructor, we can remove the field entirely and and use SparkSession.active() in onJobEnd()

Issue #1389, opened 2025-06-25
Timestamp Conversion from Julian to Gregorian Causes Date Shift in Arrow Format (0001-01-01 00:00:00 => 0001-01-03 00:00:00)


idleyui:
**Environment**
Spark Version: 3.4
Spark BigQuery Connector Version: 0.41.0

**Problem Description**
When reading a timestamp value `0001-01-01 00:00:00` from BigQuery using the Arrow format in the Spark BigQuery Connector, the result is incorrectly shifted to `0001-01-03 00:00:00`. This discrepancy does not occur when using the Avro format.

The issue arises from the use of rebaseJulianToGregorianMicros in the Arrow path ([ArrowSchemaConverter.java#L550](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/master/spark-bigquery-connector-common/third_party/apache-spark/src/main/java/com/google/cloud/spark/bigquery/ArrowSchemaConverter.java#L550)), which rebases the timestamp from the Julian calendar to the Gregorian calendar. This behavior is not present in the Avro path.

**Reproduction Steps**
1. Create Test Data in BigQuery
```sql
CREATE OR REPLACE TABLE type_timestamp (
    ts TIMESTAMP
);

INSERT INTO type_timestamp VALUES ('0001-01-01 00:00:00');
```
 
2. Spark Code (Arrow Format)

```java
Dataset<Row> df = spark.read()
    .format("bigquery")
    .option("project", "project")
    .option("parentProject", "project")
    .option("table", "project.dataset.type_timestamp")
    .load();

df.show(200, false);
df.withColumn("sec", unix_timestamp(col("ts"))).show(200, false);
```

3. Spark Code (Avro Format)
Add the following configuration to force Avro format:

```java
spark.conf().set("readDataFormat", "avro");
```

**Results**
Arrow Format (Incorrect)
```text
+-------------------+------------+
|ts                 |sec         |
+-------------------+------------+
|0001-01-03 00:00:00|-62135424343|
+-------------------+------------+
```
Avro Format (Correct)
```text
+-------------------+------------+
|ts                 |sec         |
+-------------------+------------+
|0001-01-01 00:00:00|-62135596800|
+-------------------+------------+
```

**Root Cause Analysis**
The discrepancy stems from how the Arrow and Avro formats handle timestamp conversions:

[...]

PR #1473, opened 2026-03-13
Add enableArrowTimestampRebase config option to control Julian-to-Gre…

BoqianShi:
This PR  addresses the issue mentioned in #1389:

BigQuery uses proleptic Gregorian calendar, so applying Spark's rebaseJulianToGregorianMicros() to Arrow timestamps is semantically incorrect and causes a 2-day shift for pre-Gregorian dates (e.g. 0001-01-01 becomes 0001-01-03). This adds an enableArrowTimestampRebase option (default true for backward compat) that users can set to false to disable the rebasing. Threaded through both DSV1 and DSV2 read paths.

### DIFF
--- a/README-template.md
+++ b/README-template.md
@@ -962,6 +962,14 @@ word-break:break-word
      <td>Read</td>
    </tr>
   <tr>
+    <td><code>enableArrowTimestampRebase</code>
+    </td>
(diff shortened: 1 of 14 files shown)


davidrabinowitz (changes requested):
@BoqianShi please create an integration test in spark-bigquery-connector-common/src/test/java/com/google/cloud/spark/bigquery/integration/ReadIntegrationTestBase.java that will insert a pre-gregorian data and verify the change (read it with and without the flag)

davidrabinowitz (approved):

PR #1488, opened 2026-05-15
Add timing and Job ID logging to write path operations

shameersss1:
Improve logging for write workflow by adding duration and jobId for tracking

### DIFF
--- a/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/write/context/BigQueryIndirectDataSourceWriterContext.java
+++ b/spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/write/context/BigQueryIndirectDataSourceWriterContext.java
@@ -180,10 +180,16 @@ public class BigQueryIndirectDataSourceWriterContext implements DataSourceWriter
                         config.getTableId(), schema, config.getEnableModeCheckForSchemaFields())
                     .getTableId());
         loadDataToBigQuery(sourceUris, schema);
+        long startTime = System.currentTimeMillis();
         Job queryJob =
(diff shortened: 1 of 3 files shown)


spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/write/context/BigQueryIndirectDataSourceWriterContext.java
```diff
@@ -180,10 +180,16 @@ private void commitMessages(WriterCommitMessageContext[] messages, long epochId)
                         config.getTableId(), schema, config.getEnableModeCheckForSchemaFields())
                     .getTableId());
         loadDataToBigQuery(sourceUris, schema);
+        long startTime = System.currentTimeMillis();
```
BoqianShi, line 183: I think for measuring elapsed time we usually use `System.nanoTime()`.

BoqianShi (commented):
LGTM with Minor change

spark-bigquery-connector-common/src/main/java/com/google/cloud/spark/bigquery/write/context/BigQueryIndirectDataSourceWriterContext.java
shameersss1: Although `System.nanoTime()` provides granular time, for our use case millis should be good enough for such large scale system, specially with spark most of the times are tracked at millis (listing time, io wait time etc).

BoqianShi: Ack

BoqianShi (approved):

PR #1502, opened 2026-07-13
Centralize and unify BigQuery Job telemetry and duration logging

shameersss1:
### Overview

This PR restructures how the Spark BigQuery Connector logs asynchronous job operations (Loads, Queries, Appends, Direct/Indirect Partitions). Previously, duration and completion logging was scattered, sometimes duplicated, or relied on localized client-side  System.currentTimeMillis()  implementations.

This change introduces a unified "choke point" by instrumenting the foundational wait wrappers in  BigQueryClient.java  to guarantee 100% coverage of all async operations while extracting exact cluster-side timing statistics.

  ### Key Changes
• Centralized Telemetry Injection: Modified the foundational  createAndWaitFor  and  waitForJob  methods in  BigQueryClient.java  to mandate a  String operation  identifier. Injected a  logJobTelemetry private method that polls the exact  completedJob.getStatistics().getStartTime()  &  getEndTime()  to log uniformly: BigQuery {type} job {jobId} ({operation}) completed in {duration}ms 
• Comprehensive Operation Tagging: Formally mapped all upwards callers to unambiguous strings:
      •  "Load Data from GCS to Table" 
      •  "Materialize Query to Table" 
      •  "Dynamic Partition Overwrite Commit" 
      •  "Overwrite Transaction Commit" 
      •  "Append Transaction Commit" 
      •  "Direct Query Execution" 
• Redundant Code Removal: Completely stripped the localized explicit  System.currentTimeMillis()  blocking logic that was previously implemented in the Writer Contexts since it is now natively superseded by the base wrapper.
• Stub Purging: Cleaned up the file architecture by purging completely unused legacy stubs, including the 1-arg  Job waitForJob(Job job)  and  Job createAndWaitFor(JobConfiguration.Builder) , alongside removing the now-unused  BaseServiceException  import.

### DIFF
--- a/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
+++ b/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
@@ -24,7 +24,6 @@ import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.json.gson.GsonFactory;
 import com.google.api.services.bigquery.Bigquery;
 import com.google.auth.http.HttpCredentialsAdapter;
-import com.google.cloud.BaseServiceException;
 import com.google.cloud.RetryOption;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryException;
@@ -162,7 +161,7 @@ public class BigQueryClient {
    *
    * @param job The {@code Job} to keep track of.
    */
-  public JobInfo waitForJob(Job job) {
+  public JobInfo waitForJob(Job job, String operation) {
     try {
       Job completedJob =
           job.waitFor(
@@ -178,6 +177,7 @@ public class BigQueryClient {
         throw new IllegalStateException(
             String.format("Job aborted due to timeout  : %s minutes", bigQueryJobTimeoutInMinutes));
       }
+      logJobTelemetry(completedJob, operation);
       jobCompletionListener.ifPresent(jcl -> jcl.accept(completedJob));
       return completedJob;
     } catch (InterruptedException e) {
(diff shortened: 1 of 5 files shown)


bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -178,6 +177,7 @@ public JobInfo waitForJob(Job job) {
         throw new IllegalStateException(
             String.format("Job aborted due to timeout  : %s minutes", bigQueryJobTimeoutInMinutes));
       }
+      logJobTelemetry(completedJob, operation);
```
yalimu-g, line 180: should we add a null check for completedJob?

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
```diff
@@ -162,7 +161,7 @@ public static synchronized void runCleanupJobs() {
    *
    * @param job The {@code Job} to keep track of.
    */
-  public JobInfo waitForJob(Job job) {
+  public JobInfo waitForJob(Job job, String operation) {
```
yalimu-g: Instead of pass a string, should we consider create a enum for job operation which is reusable? for example as below:
```
 public enum JobOperation {
    LOAD_DATA_FROM_GCS("Load Data from GCS to Table"),
    MATERIALIZE_QUERY("Materialize Query to Table"),
    OVERWRITE_TRANSACTION("Overwrite Transaction Commit"),
    APPEND_TRANSACTION("Append Transaction Commit"),
    DYNAMIC_PARTITION_OVERWRITE("Dynamic Partition Overwrite Commit");

    private final String description;

    JobOperation(String description) {
        this.description = description;
    }

    public String getDescription() {
        return description;
    }
}
```

shameersss1: ack

bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryClient.java
shameersss1: ack

davidrabinowitz (approved):

yalimu-g (approved):

PR #1510, opened 2026-08-06
Handle Arrow StructVectors with 0 child vectors in ArrowColumnBatchPartitionReaderContext

yalimu-g:
When reading Arrow record batches from BigQuery Storage API, there are scenarios (such as field pruning or when deserializing Arrow schema headers) where an incoming Arrow StructVector in a VectorSchemaRoot has 0 child vectors (structVector.getChildrenFromFields().isEmpty()).

When VectorLoader.load(ArrowRecordBatch) is invoked on VectorSchemaRoot, VectorLoader iterates through existing child vectors in getChildrenFromFields(). If a StructVector has 0 child vectors, VectorLoader skips loading children for that struct. However, the incoming ArrowRecordBatch still contains FieldNodes and ArrowBuf buffers for those child fields. This causes VectorLoader to get out of sync with the record batch's buffer stream, leading to buffer misalignment errors or data corruption when reading subsequent columns.

Implement helper methods ensureStructVectorsHaveChildren to traverse a VectorSchemaRoot and populate empty null child vectors for any StructVector whose getChildrenFromFields() is empty.

https://buganizer.corp.google.com/issues/540282714

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 * PR #1494: Added Support for Change Data Capture (CDC)
 * Issue #1501: Preserved detailed row errors from failed direct writes
 * Issue #1503: Corrected the documented billing project option to `parentProject`.
+* PR #1510: Handle Arrow StructVectors with 0 child vectors in ArrowColumnBatchPartitionReaderContext
 
 ## 0.44.2 - 2026-05-20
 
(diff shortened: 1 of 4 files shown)


BoqianShi (dismissed):

davidrabinowitz (changes requested):
please update the CHANGES.md file

davidrabinowitz (approved):

PR #1516, opened 2026-08-14
Build Integration Tests jar and put it under gcs during nightly job

yalimu-g:
The jars are going to be under: gs://dataproc-integration-test-input/spark-bigquery-integration-test

### DIFF
--- a/cloudbuild/nightly.sh
+++ b/cloudbuild/nightly.sh
@@ -97,12 +97,6 @@ case $STEP in
     gsutil cp "${M2REPO}/com/google/cloud/spark/spark-bigquery-metrics/${BUILD_REVISION}/spark-bigquery-metrics-${BUILD_REVISION}.jar" "gs://${BUCKET}"
     gsutil cp "gs://${BUCKET}/spark-bigquery-metrics-${BUILD_REVISION}.jar" "gs://${BUCKET}/spark-bigquery-metrics-nightly-snapshot.jar"
 
-    # Handle Spark 4.1 which has a -preview suffix in its version
-    jar_path_41="${M2REPO}/com/google/cloud/spark/spark-4.1-bigquery/${BUILD_REVISION}-preview/spark-4.1-bigquery-${BUILD_REVISION}-preview-tests.jar"
-    if [ -f "$jar_path_41" ]; then
-      gsutil cp "$jar_path_41" "gs://dataproc-integration-test-input/spark-bigquery-integration-test/spark-4.1-bigquery-latest-tests.jar"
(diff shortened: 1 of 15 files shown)


cloudbuild/nightly.sh
```diff
@@ -97,6 +97,9 @@ case $STEP in
     gsutil cp "${M2REPO}/com/google/cloud/spark/spark-bigquery-metrics/${BUILD_REVISION}/spark-bigquery-metrics-${BUILD_REVISION}.jar" "gs://${BUCKET}"
     gsutil cp "gs://${BUCKET}/spark-bigquery-metrics-${BUILD_REVISION}.jar" "gs://${BUCKET}/spark-bigquery-metrics-nightly-snapshot.jar"
 
+    # Upload the newly created integration tests shaded jar
+    find "${M2REPO}/com/google/cloud/spark" -name "*-tests.jar" -exec gsutil cp {} "gs://dataproc-integration-test-input/spark-bigquery-integration-test/" \;
```
davidrabinowitz: please replace `gsutil` with `gcloud storage`

davidrabinowitz (changes requested):
how many test jars are we going to have? The tests are all in spark-bigquery-connector-common, any reason to have more than one jar?

cloudbuild/nightly.sh
yalimu-g: should I replace all the above gsutil?

yalimu-g:
> how many test jars are we going to have? The tests are all in spark-bigquery-connector-common, any reason to have more than one jar?

If we want to test it against several versions of cluster and s8s, should we build it against different spark and Scala versions?

davidrabinowitz:
> > how many test jars are we going to have? The tests are all in spark-bigquery-connector-common, any reason to have more than one jar?
> 
> If we want to test it against several versions of cluster and s8s, should we build it against different spark and Scala versions?

it is written in Java, and the spark APIs are backwards compatible. Even when used in memory, the same spark-bigquery-connector-common is used in the classpath

yalimu-g:
> > > how many test jars are we going to have? The tests are all in spark-bigquery-connector-common, any reason to have more than one jar?
> > 
> > 
> > If we want to test it against several versions of cluster and s8s, should we build it against different spark and Scala versions?
> 
> it is written in Java, and the spark APIs are backwards compatible. Even when used in memory, the same spark-bigquery-connector-common is used in the classpath

I think the idea is to test the same version using integration tests as the spark-bq-connector jar. You can see the in progress reference here cl/963710916

PR #1522, opened 2026-09-18
Report a clear error when reading a BigQuery table that does not exist (DataSource V2)

BoqianShi:
### Issue

Reading a BigQuery table that does not exist with the DataSource V2 connector failed with a `NullPointerException` instead of reporting the missing table:

```python
spark.read.format("bigquery").option("table", "my-project.my_dataset.no_such_table").load()
```

```text
org.apache.spark.SparkException: [INTERNAL_ERROR] Eagerly executed command failed.
Caused by: java.lang.NullPointerException: Cannot invoke
  "org.apache.spark.sql.types.StructType.fields()" because "schema" is null
```

The same read with DataSource V1 already reported `Table <name> not found`.

### Fix

The DataSource V2 connector now reports the same message as DataSource V1:

```text
Table my-project.my_dataset.no_such_table not found
```

Writing a DataFrame to a table that does not exist yet is unchanged — that path still creates the table.

### Tests

* New unit tests for the schema lookup and for the table provider's behavior when the table is missing.
* `ReadIntegrationTestBase.testNonExistentSchema` now asserts the error message. It previously accepted any `RuntimeException`, so it passed on the `NullPointerException` above.


spark-bigquery-dsv2/spark-3.1-bigquery-lib/src/main/java/com/google/cloud/spark/bigquery/v2/Spark3Util.java
```diff
@@ -53,16 +55,27 @@ public static Table createBigQueryTableInstance(
...
+    Schema schemaFromTable = bigQueryClient.getReadTableSchema(config.toReadTableOptions());
+    if (schemaFromTable == null) {
+      throw new TableNotFoundException(
+          "Table " + BigQueryUtil.friendlyTableName(config.getTableId()) + " not found");
+    }
+    return SchemaConverters.from(SchemaConvertersConfiguration.from(config))
+        .toSpark(schemaFromTable);
+  }
+
+  static class TableNotFoundException extends BigQueryConnectorException {
```
davidrabinowitz: We can use the built in org.apache.spark.sql.catalyst.analysis.NoSuchTableException

BoqianShi: `NoSuchTableException` is a checked exception because it extends `AnalysisException`, which extends `Exception`. Our schema lookup runs inside a `Supplier<StructType>`, so replacing the current exception with it would fail compilation:

```java
Supplier<StructType> schemaSupplier =
    () -> { throw new NoSuchTableException("dataset", "table"); };
```

```text
unreported exception NoSuchTableException;
must be caught or declared to be thrown
```

Neither `Supplier.get()` nor Spark’s `Table.schema()` declares checked exceptions, and Java does not allow our implementation to add them. I verified this across Spark 3.1–4.1. We could wrap the native exception in an unchecked exception, but that would still require a wrapper and additional cause handling; I suggest keeping the dedicated unchecked exception so `inferSchema()` can catch missing-table failures specifically and preserve the write fallback.

davidrabinowitz: Ok, please just extract TableNotFoundException to its own file

davidrabinowitz (changes requested):

spark-bigquery-dsv2/spark-3.1-bigquery-lib/src/main/java/com/google/cloud/spark/bigquery/v2/Spark3Util.java
BoqianShi: Done

